PR #58, opened 2015-01-27
Added an example for StereoPannerNode

hoch:
Added an example for StereoPannerNode demoing basic functionality. Also a new LICENSE file and a base style sheet for future improvement were added.


hoch:
To address this issue #59, I added a small feature detection routine. If UA doesn't have a certain feature in AudioContext, an error message will be thrown and the demo page will be locked from user interaction.

PR #62, opened 2015-05-14
Added link to sample index page

cmumford:


hoch:
We already have a working URL for the index page. It seems to be redundant to have another (which is identical) and the existing URL is shorter and more readable.

http://googlechrome.github.io/web-audio-samples/

Can you clarify your intention?

cmumford:
Silly me - going too quickly. I thought those were links to the old project locations before moving to GitHub.

PR #66, opened 2015-07-30
Fixed to use AudioContext

Korilakkuma:
webkitAudioContext is deprecated.


rtoy:
Thanks for these fixes.  I think we fixed earlier most of the important demos to support both prefixed and un-prefixed AudioContext.  Nice to have all of the remaining demos fixed, even if we don't really know if they work or not, or even what they're supposed to do.

PR #69, opened 2015-09-03
Update timeline demo

rtoy:
Primarily adds an example of setTargetAtTime.

Secondarily, replace the old graph code with flot which produces a nicer looking graph. Can't save the graph as an image anymore, but a screenshot works just as well.


rtoy:
I signed it!

rtoy:
Closing this.  This is a duplicate of #68.

PR #73, opened 2017-03-28
Explicit cleanup of finished audio nodes

bjornm:
Added explicit cleanup of finished audio nodes. This avoids gradual performance degradation due to lazy garbage collection. By explicitly disconnecting ABSN and other voice-related nodes from the audio graph once they're finished, we can avoid the situation where CPU load keeps increasing although number of playing nodes do not.

If you click garbage collect in the dev tools some of the CPU load goes away so this is probably related to Chrome web audio internal garbage collection.


bjornm:
I signed it!

Björn

PR #82, opened 2017-06-16
Organize the index page.

hoch:
I put the quick link section to our favorite demos. We have multiple 'index' pages here and there, but this index should be the entry point for everything. I intend to continue to work on this.

### DIFF
--- a/index.html
+++ b/index.html
@@ -24,9 +24,14 @@
         </a>
       </li>
       <li>
+        <a href="https://hoch.github.io/canopy/">Canopy</a>
+      </li>
+      <li>
         <a href="https://github.com/GoogleChrome/web-audio-samples">Chrome
         WebAudio Sample GitHub repo</a>
       </li>
+    </ul>
+    <ul>
       <li>
         <a href="samples/audio/index.html">Sample index page (circa 2010)</a>
       </li>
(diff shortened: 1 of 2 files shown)


resources/base.css
```diff
@@ -0,0 +1,4 @@
+body {
+  font-family: "Roboto", sans-serif;
```
rtoy: What happens if Roboto isn't available?

index.html
```diff
@@ -1,64 +1,235 @@
...
+        <a href="samples/audio/index.html">Sample index page (circa 2010)</a>
+      </li>
+      <li>
+        <a href="samples/audio/samples.html">Third-party samples</a>
+      </li>
+      <li>
+        <a href="stress-test/boxes">Box2D Stress Test</a>
+      </li>
+      <li>
+        <a href="https://hoch.github.io/canopy/">Canopy</a>
```
rtoy: I think canopy should be right after the spec link.

Possibly separate out the spec link and canopy from the other items into it's own list. Kind of like:

- spec
- canopy
- github 

[blank line]
- old index
- 3rd party
- stress test

resources/base.css
hoch: Then fall back to sans-serif, which is universally available.

PR #84, opened 2017-06-20
Noise gate

samhincks:
The main algorithm for the noise gate without test cases or a GUI.


samhincks:
That's an error. There should only be one file in it. I will delete this and open a new one.

samhincks:
Looking into it some more, I think these got mixed up since I did git branch noise-gate while I was in the the sam branch and not from the master branch. Since I only staged and committed the noisegate file.

samhincks:
I don't think I have the authority to close / delete this pull request. But I opened a new one which only has the one file for the actual noise gate in it.

samhincks:
Post mortem on this: I'm pretty sure it occurred since I had pulled files from the sam branch while in the noise gate branch. But I'm still not 100% sure.

rtoy:
Closing.

PR #86, opened 2017-06-27
Adding `audio-worklet/basic` with basic demos

hoch:
PTAL.


audio-worklet/basic/audio-param-modulation.html
```diff
@@ -1,41 +1,40 @@
...
+    <script src="../lib/audio-worklet-helper.js"></script>
+  </head>
+  <body>
+    <a href="../index.html" class="link-small">&#171; AudioWorklet home</a>
+    <h1>
+      Demo: AudioParam Modulation
+    </h1>
+    <script>
+    // TODO(hoch): remove this GC hack when it is implemented properly.
+    let paramAmp;
```
rtoy: Does this mean that GC removes paramAmp even though it's in use?

audio-worklet/basic/js/bit-crusher.js
```diff
@@ -1,40 +1,46 @@
...
-      maxValue: 16 
-    }, {
-      name: 'frequencyReduction',
-      defaultValue: 0.5,
-      minValue: 0,
-      maxValue: 1
-    }];
+  static get parameterDescriptors() {
+    return [
+      {name: 'bitDepth', defaultValue: 12, minValue: 1, maxValue: 16}, {
```
rtoy: Is this clang-format?  If so, I really, Really, REALLY hate how this is formatted.

audio-worklet/basic/audio-param-modulation.html
hoch: That's what the current prototype does.

audio-worklet/basic/js/bit-crusher.js
hoch: Yeap. Well, we decided to bite the bullet. Didn't we? :)

hoch:
Actually I made changes in `.gitignore` to exclude `.DS_Store`. So this PR is removing them from the repository.

audio-worklet/basic/js/bit-crusher.js
rtoy: Yes, we did. I just wish clang-format had some consistency.  If some item is spread out over several lines, do it for all.  Or not.  Just not some random mix.

But that's life, so this is fine.

PR #87, opened 2017-07-15
Bitcrusher worklet

samhincks:
The bitcrusher worklet demo and modifications to directory structure.


samhincks:
Check out the demo at https://rawgit.com/GoogleChrome/web-audio-samples/dee6ac60bb9001ce72d96aec7e48c033b6a17907/audio-worklet/bitcrusher.html 

But you need to be using the 07-12 custom build of Chromium for it to work.

audio-worklet/bitcrusher.html
```diff
@@ -13,46 +13,48 @@
...
-      });
+    <script>
+      function runBitcrusherDemo() {
+          window.audioWorklet.addModule('src/bitcrusher-audio-worklet.js')
+              .then(() => {
+                  const context = new AudioContext();
+                  let demo = new BitcrusherDemo(context, 'container');
+              });
+      }
+      AudioWorkletHelper.startDemo(runBitcrusherDemo);
```
rtoy: Doesn't this mean that you can't run the script processor demo anymore.  Is this really what we want?  I was kind of expecting to be able to run both (not at the same time of course).

audio-worklet/src/bitcrusher-audio-worklet.js
```diff
@@ -0,0 +1,73 @@
...
+      defaultValue: 1,
+      minValue: 1,
+      maxValue: 20
+    }];
+  }
+
+  constructor(options) {
+    super(options);
+
+    // Index and previousSample defined as globals to handle block transitions.
```
rtoy: |index_| and |previousSample_|  in comment?

audio-worklet/src/bitcrusher-audio-worklet.js
```diff
@@ -0,0 +1,73 @@
...
+   * @param {Float32Array} parameters.reduction sample rate reduction
+   *                                            for each sample
+   */
+  process(input, output, parameters) {
+    let inputChannelData = input.getChannelData(0);
+    let outputChannelData = output.getChannelData(0);
+    
+    // Add new bit crushed sample to outputBuffer at specified interval.
+    for (let j = 0; j < inputChannelData.length; j++) {
+      if (this.index_ % parameters.reduction[j] === 0) {
```
rtoy: This kind of implies that you're doing a-rate processing for the reduction.  Is this intended?

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -16,24 +16,52 @@
...
+    // GUI default settings depend on parameter values defined in audio-worklet.
+    this.bitDepthMaxValue_ = this.paramBitDepth_.maxValue;
+    this.bitDepthMinValue_ = this.paramBitDepth_.minValue;
+    this.bitDepthDefaultValue_ = this.paramBitDepth_.defaultValue;
+    this.reductionMaxValue_ = this.paramReduction_.maxValue;
+    this.reductionMinValue_ = this.paramReduction_.minValue;
+    this.reductionDefaultValue_ = this.paramReduction_.defaultValue;
+
+    // For the audio worklet, the bit depth and reduction parameters are
+    // modulated by changing the offset to connected constant source nodes.
```
rtoy: Why?  I was expecting direct manipulation of the AudioParams of the worklet.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -16,24 +16,52 @@
...
+    this.bitDepthOffsetNode_ = new ConstantSourceNode(context, {offset:0});
+    this.reductionOffsetNode_ = new ConstantSourceNode(context, {offset:0});
+    this.bitDepthOffsetNode_.connect(this.paramBitDepth_);
+    this.reductionOffsetNode_.connect(this.paramReduction_);
+    this.bitDepthOffsetNode_.start();
+    this.reductionOffsetNode_.start();
 
-    this.bitcrusher_.output.connect(this.masterGain_);
+    // The user can swap between the bitcrusher and audio worklet node
```
rtoy: You mean between ScriptProcessor and AudioWorklet, not bitcrusher and audio worklet.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -16,24 +16,52 @@
...
+    // by checking and unchecking useAudioWorkletCheckBox_.
+    this.bitcrusherScriptProcessor_ = new Bitcrusher(this.context_, {
+        channelCount: 1,
+        bitDepth: this.bitDepthDefaultValue_,
+        reduction: this.reductionDefaultValue_
+    });
+    
+    // The audio worklet bitcrusher is used by default.
```
rtoy: What if it's not available?  People will run this on implementations that will not have it.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,16 +113,50 @@ class BitcrusherDemo {
           default: this.masterGain_.gain.value,
           name: 'Volume'
         });
+
+    // GUI controls allow switching between audio worklet and script processor.
+    // The audio worklet node is used by default.
+    this.useAudioWorkletCheckBox_ = document.createElement('input');
+    this.useAudioWorkletCheckBox_.type = 'checkbox';
```
rtoy: Why not a radio button?  I'm assuming the demo must either use the script OR the worklet, not both, so two buttons would seem more natural to select which mode you want to try.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,16 +113,50 @@ class BitcrusherDemo {
...
+    this.useAudioWorkletCheckBox_.checked = true;
```
rtoy: No, you can't make this the default because the AudioWorklet may not be available.  You have to check if it's available first.

I would also have the worklet checkbox disabled if worklets aren't available.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,16 +113,50 @@ class BitcrusherDemo {
...
+    const audioworkletLabel = document.createElement('label');
+    audioworkletLabel.textContent =
+        'Use Audio Worklet (if unchecked, use Script Processor Bitcrusher).';
```
rtoy: Radio buttons would take care of this, right?

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,16 +113,50 @@ class BitcrusherDemo {
...
+    container.appendChild(audioworkletLabel);
+  }
+
+  /**
+   * Switch between whether the audio worklet node or script processor node
+   * connects to the master gain node.
+   */
+  switchBitcrusherNode() {
+    if (this.useAudioWorkletCheckBox_.checked) {
+       this.bitcrusherScriptProcessor_.output.connect(this.masterGain_);
```
rtoy: Does it really make sense in the demo to be able to dynamically switch between the script processor and the worklet?  At the very least the delays are different so it will sound funny.

I was kind of expecting the user to pick one mode and not dynamically switching while the music is playing.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,16 +113,50 @@ class BitcrusherDemo {
...
-   * This bound to event listener by a ParamController.
+   * This is bound to an event listener by a ParamController.
    * @param {Number} value the new bit depth
    */
   setBitDepth(value) {
     if (value < 1) console.error('The minimum bit depth rate is 1.');
-    this.bitcrusher_.bitDepth = value;
+    
+    this.bitDepthOffsetNode_.offset.value =
+          value - this.bitDepthDefaultValue_;
```
rtoy: Why are you subtracting here?

audio-worklet/src/bitcrusher-audio-worklet.js
samhincks: @hoch told me to use a-rate for both parameters. I'm guessing only a-rate works for now?

audio-worklet/src/bitcrusher-demo.js
samhincks: My mistake. I thought that since the params needed to be driven by a constant source node they also needed to be modified by that constant source node, but that is not the case.

samhincks: Oh turns out they don't even need to be driven by the constant source nodes anymore, so I guess that changed from the previous chromium build I had

audio-worklet/src/bitcrusher-demo.js
samhincks: I think this is what I discussed with @hoch

audio-worklet/src/bitcrusher-demo.js
hoch: Yeap. The build 0712 now supports all AudioParam operations. You can take ConstantSources out of this demo.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -16,24 +16,52 @@
...
+    this.bitDepthOffsetNode_ = new ConstantSourceNode(context, {offset:0});
+    this.reductionOffsetNode_ = new ConstantSourceNode(context, {offset:0});
+    this.bitDepthOffsetNode_.connect(this.paramBitDepth_);
+    this.reductionOffsetNode_.connect(this.paramReduction_);
+    this.bitDepthOffsetNode_.start();
+    this.reductionOffsetNode_.start();
 
-    this.bitcrusher_.output.connect(this.masterGain_);
+    // The user can swap between the bitcrusher and audio worklet node
+    // by checking and unchecking useAudioWorkletCheckBox_.
```
hoch: Use a toggle button instead of checkbox.

audio-worklet/src/bitcrusher-demo.js
hoch: Audio Worklet -> AudioWorklet
Script Processor -> ScriptProcessor

audio-worklet/src/bitcrusher-demo.js
hoch: We discussed together and agreed the on-the-fly A-Bing makes sense.

audio-worklet/src/bitcrusher-demo.js
hoch: I don't think this is relevant anymore since ConstantSourceNode will be removed.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -104,7 +166,11 @@ class BitcrusherDemo {
    */
   setReduction(value) {
     if (value < 1) console.error('The minimum reduction rate is 1.');
```
hoch: Let's be consistent. `throw` or `console.error`?

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -123,13 +189,16 @@ class BitcrusherDemo {
     // Play song, running samples through a bitcrusher under user control.
     this.song_ =
```
hoch: It's hard to refer `song_` below. Please rename it something implies `bufferSource`.

audio-worklet/src/bitcrusher-audio-worklet.js
rtoy: Couple of issues here, I think.  First, which should happen first?  Bit reduction or sample rate reduction?  You get different values depending on which is done first.  This is true for the script processor version too.

Second, this makes it different from the script processor.

Third, if reduction is not an integer, line 62 will almost never be true.  Now what?

Because it's an audio param, things are a bit more complicated because you can connect an arbitrary signal to control the sample rate reduction.

Not sure if we want to fix this.  Or maybe just add a comment that this needs to be fixed.

audio-worklet/src/bitcrusher-demo.js
samhincks: They receive this error: [ERROR] The browser does not support AudioWorklet yet.
Which is caught in lib/audio-worklet-helper.js

samhincks:
Demo is at https://rawgit.com/GoogleChrome/web-audio-samples/725ea9503f1b86fe6cffaad0ceea0931d70b0664/audio-worklet/bitcrusher.html
Use this one instead of the last link, since the GUI wasn't appropriately reflecting whether script processor or audio worklet was being used

audio-worklet/src/bitcrusher-audio-worklet.js
rtoy: My mistake.  The order of reduction and bit crushing is fine.  But you still need to figure out what to do with non-integer reduction values because the code as is doesn't really support that.

audio-worklet/src/bitcrusher-audio-worklet.js
```diff
@@ -39,7 +39,7 @@ registerProcessor('bitcrusher-audio-worklet',
   constructor(options) {
     super(options);
 
-    // Index and previousSample defined as globals to handle block transitions.
+    // Index_ and previousSample_ are globals to handle block transitions.
```
rtoy: Nit: The variable is `index_` and not `Index_`.  I think this is one reason people do `|index_|`.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -31,35 +31,22 @@ class BitcrusherDemo {
...
-    this.reductionOffsetNode_ = new ConstantSourceNode(context, {offset:0});
-    this.bitDepthOffsetNode_.connect(this.paramBitDepth_);
-    this.reductionOffsetNode_.connect(this.paramReduction_);
-    this.bitDepthOffsetNode_.start();
-    this.reductionOffsetNode_.start();
-
-    // The user can swap between the bitcrusher and audio worklet node
+    // The script processor will have a delay relative to the audio worklet node
+    // proportional to the size of its buffersize. This delay can be heard as
+    // a glitch as the user switches node.
```
rtoy: The comment about glitching isn't really relevant to the description of this variable.  But oh well.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,16 +100,49 @@ class BitcrusherDemo {
...
+    container.appendChild(audioworkletLabel);
+  }
+
+  /**
+   * Switch between whether the audio worklet node or script processor node
+   * connects to the master gain node.
+   */
+  switchBitcrusherNode() {
+    // The function is called after the checkbox switches state.
+    const changeToScriptProcessor = !this.useAudioWorkletCheckBox_.checked;
```
rtoy: Why invert the logic?

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -16,24 +16,39 @@
...
+    this.bitcrusherAudioWorklet_ =
+        new AudioWorkletNode(this.context_, 'bitcrusher-audio-worklet');
+    this.paramBitDepth_ =
+        this.bitcrusherAudioWorklet_.parameters.get('bitDepth');
+    this.paramReduction_ =
+        this.bitcrusherAudioWorklet_.parameters.get('reduction');
+
+    // The script processor will have a delay relative to the audio worklet node
+    // proportional to the size of its buffersize. This delay can be heard as
+    // a glitch as the user switches node.
```
hoch: Well, I am not sure if I can understand this comment. What do 'relative' and 'proportional' mean here w.r.t AudioWorkletNode?

audio-worklet/test/bitcrusher-test.html
```diff
@@ -14,7 +14,7 @@
 -->
 <!DOCTYPE html>
```
hoch: The custom build does not support OfflineAudioContext yet, so I think you can remove this test file out of this PR.

audio-worklet/bitcrusher.html
hoch: I think this template code needs to be rewrite. It was originally designed for the AudioWorklet demo.

audio-worklet/src/bitcrusher-demo.js
samhincks: It's inverted since the checkbox will have unchecked itself before the switchBitcrusherNode is called from the event listener.

samhincks:
Here is the link to the new bitcrusher which works even if the browser does not support script processor: https://rawgit.com/GoogleChrome/web-audio-samples/f94f55a9f2dd780af5763aa03cc908d28d909c1f/audio-worklet/bitcrusher.html

The implementation does not use radio buttons since the code would require these to be added programmatically which is complex - but I can make this change if needed.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -68,6 +57,16 @@ class BitcrusherDemo {
       this.reductionMin_ = this.paramReduction_.minValue;
     }
 
+    // The script processor will have a delay proportional to the size of its
+    // buffersize, which can be heard as a glitch as the user switches node.
```
rtoy: Don't see how this comment about glitching is relevant in any way to the buffer size variable. I think it would be more appropriate to explain this in the text for the demo.  So the user knows a glitch is possible.

audio-worklet/test/bitcrusher-test.html
hoch: If you want to keep this file, please add a comment that says this will work when AudioWorkletNode works on OfflineAudioContext.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -102,20 +125,23 @@ class BitcrusherDemo {
...
-
-    this.useAudioWorkletCheckBox_.addEventListener(
-        'change', this.switchBitcrusherNode.bind(this));
-
-    let container = document.getElementById(containerId);
-    container.appendChild(this.useAudioWorkletCheckBox_);
-    container.appendChild(audioworkletLabel);
+    // The script processor node is used by default, and switching is not
+    // enabled if the browser does not support audio worklet.
+    if (this.useWorkletAndScriptProcessor_) {
```
rtoy: So, if AudioWorklet isn't available, you don't even show the checkbox?  Is that the right UI? I was kind of expecting the checkbox to be there, but unclickable.  So the user knows something is missing.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -124,13 +150,13 @@ class BitcrusherDemo {
    */
   switchBitcrusherNode() {
     // The function is called after the checkbox switches state.
-    const changeToScriptProcessor = !this.useAudioWorkletCheckBox_.checked;
-    if (changeToScriptProcessor) {
-      this.bitcrusherScriptProcessor_.output.connect(this.masterGain_);
-      this.bitcrusherAudioWorklet_.disconnect();
-    } else {
+    const changeToAudioWorklet = !this.useScriptProcessorCheckBox_.checked;
```
rtoy: You didn't answer this from last time.  Why the convoluted logic where you invert useScriptProcessorCheckbox_.checked?

audio-worklet/bitcrusher.html
```diff
@@ -13,46 +13,58 @@
...
+      // switch between audio worklet and script processor implementations of
+      // the bitcrusher.
+      const useAudioWorkletAndScriptProcessor = window.audioWorklet &&
+          typeof window.audioWorklet.addModule === 'function';
 
-      window.addEventListener('load', () => {
-          let controller = new BitcrusherDemo(
-              'container', bitDepth, reduction, gain);
-      });
+      if (useAudioWorkletAndScriptProcessor) {
```
hoch: Let's just say you isolate this variable from the context. What does it mean that `useAudioWorkletAndScriptProcessor` being false? Using AudioWorkletNode only or the other way around?

I think the boolean flag should be named with a clear intention.

I can see this one boolean flag is affecting all of your switching logic below. Please reconsider your approach. I will the review the rest of demo when you address this.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,26 +123,65 @@ class BitcrusherDemo {
...
+      container.appendChild(this.useScriptProcessorCheckBox_);
+      container.appendChild(audioworkletLabel);
+    }
+  }
+
+  /**
+   * Switch between whether the audio worklet node or script processor node
+   * connects to the master gain node.
+   */
+  switchBitcrusherNode() {
```
hoch: `toggle` is a better name.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,26 +123,65 @@ class BitcrusherDemo {
...
+    // The function is called after the checkbox switches state.
+    const changeToAudioWorklet = !this.useScriptProcessorCheckBox_.checked;
```
hoch: So no radio button?

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -124,13 +150,13 @@ class BitcrusherDemo {
...
-      this.bitcrusherScriptProcessor_.output.connect(this.masterGain_);
-      this.bitcrusherAudioWorklet_.disconnect();
-    } else {
+    const changeToAudioWorklet = !this.useScriptProcessorCheckBox_.checked;
+    if (changeToAudioWorklet) {
       this.bitcrusherAudioWorklet_.connect(this.masterGain_);
       this.bitcrusherScriptProcessor_.output.disconnect();
+    } else {
+      this.bitcrusherScriptProcessor_.output.connect(this.masterGain_);
+      this.bitcrusherAudioWorklet_.disconnect();
```
rtoy: I thought we had discussed the switching would be done with a pair of gain nodes.  This connect/disconnect isn't sample-accurate.  Plus you connect before disconnect, so there's a chance you get output from both at the same time.

With gain nodes and AudioParam control, you can change them at the same time guaranteed (if you schedule a bit in the future).

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -85,26 +123,65 @@ class BitcrusherDemo {
...
+  switchBitcrusherNode() {
+    // The function is called after the checkbox switches state.
+    const changeToAudioWorklet = !this.useScriptProcessorCheckBox_.checked;
+    if (changeToAudioWorklet) {
+      this.bitcrusherAudioWorklet_.connect(this.masterGain_);
+      this.bitcrusherScriptProcessor_.output.disconnect();
+    } else {
+      this.bitcrusherScriptProcessor_.output.connect(this.masterGain_);
+      this.bitcrusherAudioWorklet_.disconnect();
+    }
```
hoch: If the AudioWorklet is not activated, this method should not do anything at all.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -141,18 +167,20 @@ class BitcrusherDemo {
    */
   setBitDepth(value) {
     if (value < 1) throw 'The minimum bit depth rate is 1.';
-    this.paramBitDepth_.value = value;
+    if (this.useWorkletAndScriptProcessor_)
+      this.paramBitDepth_.value = parseFloat(value);
```
rtoy: Why do you need a `parseFloat` here but not for setting the script processor bit depth?

Similarly below for setting the reduction value.

audio-worklet/src/bitcrusher-audio-worklet.js
```diff
@@ -61,8 +62,7 @@ registerProcessor('bitcrusher-audio-worklet',
     
     // Add new bit crushed sample to outputBuffer at specified interval.
     for (let j = 0; j < inputChannelData.length; j++) {
-      if (this.index_ % parameters.reduction[j] === 0) {
-        
+      if (this.index_ % Math.round(parameters.reduction[j]) === 0) {
```
rtoy: Add a comment that only integer reduction is supported, hence the rounding.  This should probably also be mentioned in the documentation for the NoiseGate as well.

audio-worklet/src/bitcrusher-audio-worklet.js
```diff
@@ -40,7 +40,8 @@ registerProcessor('bitcrusher-audio-worklet',
   constructor(options) {
     super(options);
 
-    // Index_ and previousSample_ are globals to handle block transitions.
+    // Index_ and previousSample_ are member variables to handle block
```
rtoy: In a previous review I said the variable is `index_` and not `Index_`.  Perhaps use `|index_|` to make it clear?

`index_` isn't a great variable name and the comment doesn't help, but ok.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -18,36 +18,59 @@ class BitcrusherDemo {
    * Initalizes and modifies bitcrusher settings in response to GUI input.
    * @param {AudioContext} context the Audio Context
    * @param {String} containerId container for GUI elements
+   * @param {Boolean} useWorkletAndScriptProcessor switch depending on browser
+   *                  which determines if both script processor and audio
+   *                  worklet should be used. If false, only the script
+   *                  processor will be used.
    */
-  constructor(context, containerId) {
+  constructor(context, containerId, useWorkletAndScriptProcessor) {
```
rtoy: I think `useWorkletAndScriptProcessor` is a bit confusing. The script processor is always used, so this really means use a worklet, right?

audio-worklet/src/bitcrusher-demo.js
samhincks: I answered it the previous time you brought it up. It's because the event listener calls the function after it has switched boolean value.

rtoy: Sorry. The review system is really convoluted because I didn't see the message in the review stream.

Still, you could have
```
const changeToScript = this.useScriptProcessorCheckBox_.checked;
if (change) {
  // script stuff
} else {
  // worklet stuff
}
```
Isn't this equivalent?

audio-worklet/src/bitcrusher-demo.js
samhincks: I can add the radio button if needed but @rtoy said the checkbox would be fine. The issue with the radio button is that it involves a lot of code to add it programmatically, and I can't see how to achieve equivalent results by hardcoding it into the HTML

audio-worklet/src/bitcrusher-demo.js
samhincks: With current design, the checkbox is only added to the GUI if audio worklet is implemented in the browser, so the method is never called.

audio-worklet/src/bitcrusher-demo.js
samhincks: I had thought you meant we would be switching between the node and the masterGain, but it sounds like you are saying that the script processor and audio worklet each have dedicated gain nodes and we switch between their respective gain nodes and the masterGain? For the script processor that would mean three gain nodes separating it from the destination since its output node is a gain node.

audio-worklet/src/bitcrusher-demo.js
rtoy: Why is it programatically hard?  Just add a an `<input>` section with two buttons.  Programatically disable one if worklets aren't available so you can't select it. (I think that's what I did for my IIR fitler design that currently doesn't support bandpass/bandstop filters.)

In any case I will leave it up to you on whether to have a checkbox or radio buttons.

But I definitely want to see a checkbox even if worklets aren't available.  It's a sign to the user that some feature is missing.

audio-worklet/src/bitcrusher-demo.js
rtoy: As I said, you connect before disconnect.  There's a possibility that you'll have both going to the output because connect and disconnect aren't sample accurate.  If you disconnect before connect, there's a chance where nothing is connected causing yet another glitch. Of course, there will be a glitch anyway since the delays aren't the same between the two paths, but no need to introduce more sources of glitching.

Gain nodes with automation will fix this.

And does it really matter there are extra gain nodes?  It shouldn't.

samhincks:
New link can be found at https://rawgit.com/GoogleChrome/web-audio-samples/bcaeed3b89ceeaf7962e4f093fb2b072dd2cebac/audio-worklet/bitcrusher.html 

(Note that the only change made to the test file is a comment saying it will work with audio worklet once offline audio context is supported)

audio-worklet/bitcrusher.html
```diff
@@ -45,25 +45,37 @@
       of distinct samples in the signal. Increasing this value simulates
       sample rate reduction by making intervals of consecutive samples equal.
     </p>
+    <p>
+      If your browser supports the AudioWorklet node, you can switch between 
+      the ScriptProcessor and AudioWorklet Bitcrusher implementations,
+      but if switching occurs as audio is played, there will be an audible
```
rtoy: Nit: "will be" -> "may be"

audio-worklet/src/bitcrusher-audio-worklet.js
```diff
@@ -40,15 +40,16 @@ registerProcessor('bitcrusher-audio-worklet',
   constructor(options) {
     super(options);
 
-    // Index_ and previousSample_ are member variables to handle block
-    // transitions.
+    // index_ and previousSample_ are member variables to handle block
+    // transitions. index_ refers to the current position in the input
+    // AudioBuffer, and previousSample_ refers to the corresponding sample.
```
rtoy: Nit: Renaming `index_` to `currentInputPosition_` would be nice because it explains what it really is.

But ok, if you leave it as is.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -17,16 +17,15 @@ class BitcrusherDemo {
   /**
    * Initalizes and modifies bitcrusher settings in response to GUI input.
    * @param {AudioContext} context the Audio Context
-   * @param {String} containerId container for GUI elements
-   * @param {Boolean} useWorkletAndScriptProcessor switch depending on browser
+   * @param {Boolean} useWorklet switch depending on browser
```
rtoy: Nit:  `useWorklet`  implies to me that you're only using a worklet.  But in fact it's more like `enableWorket` so that you can use a worklet but you don't have to.  Which is the case in this demo.

The comments describe the purpose of this clearly, but having a name that reflects that is better.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -124,40 +129,26 @@ class BitcrusherDemo {
...
-          'change', this.switchBitcrusherNode.bind(this));
-
-      let container = document.getElementById(containerId);
-      container.appendChild(this.useScriptProcessorCheckBox_);
-      container.appendChild(audioworkletLabel);
-    }
+    let workletButton = document.getElementById(workletButtonId);
+    if (this.useWorklet_)
+      workletButton.addEventListener('click', this.workletSelected.bind(this));
+    else workletButton.disabled = true;
```
rtoy: Should this be on two lines, with `else` all by itself?

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -124,40 +129,26 @@ class BitcrusherDemo {
...
-      this.bitcrusherScriptProcessor_.output.disconnect();
-    } else {
-      this.bitcrusherScriptProcessor_.output.connect(this.masterGain_);
-      this.bitcrusherAudioWorklet_.disconnect();
-    }
+  workletSelected() {
+    // Switching occurs 10ms into the future for thread synchronization.
+    this.workletGain_.gain.setValueAtTime(1, this.context_.currentTime + 0.01);
+    this.scriptProcessorGain_.gain.setValueAtTime(
+        0, this.context_.currentTime + 0.01);
```
rtoy: Because `currentTime` can advance between the two calls to `setValueAtTime`, you probably want to do something like:
```
let t = context.currentTime + 0.01;
workletGain.gain.setValueAtTime(1, t);
scriptGain.gain.setValueAtTime(0,t);
```

audio-worklet/src/bitcrusher-audio-worklet.js
samhincks: I'm going to leave it as it is so that it has a clear correspondance to scriptprocessor implementation.

audio-worklet/bitcrusher-test.html
```diff
@@ -14,17 +14,20 @@
...
 
       /**
        * Verify the sample rate reduction component of the bitcrusher operating
-       * outside of an audio context. A bitcrushed array generated from an 
-       * arbitrary sequence of numbers should hold sequences of consecutive 
+       * outside of an audio context. A bitcrushed array generated from an
+       * arbitrary sequence of numbers should hold sequences of consecutive
        * equal numbers, controlled by the variable factor.
        */
       function testSampleReduction() {
```
hoch: `testSampleRateReduction()`

audio-worklet/bitcrusher-test.html
```diff
@@ -48,34 +51,34 @@
...
           });
         }
       }
 
       /**
-       * Verify that the bitcrusher does not alter samples beyond an acceptable 
+       * Verify that the bitcrusher does not alter samples beyond an acceptable
        * margin when parameterized to avoid information reduction.
        */
       function testBitcrusherInAudioContext() {
```
hoch: `testBitcrusherInOfflineAudioContext()`

audio-worklet/bitcrusher-test.html
```diff
@@ -48,34 +51,34 @@
...
             let expected = Math.floor(index / reduction) * reduction;
             console.assert(
-                data === expected, 
+                data === expected,
                 'computed ' + data + ' but expected ' + expected);
           });
         }
       }
 
       /**
```
hoch: Use `//` for the generic comment. `/**...*/` is reserved for JSDoc.

audio-worklet/bitcrusher-test.html
```diff
@@ -48,34 +51,34 @@
...
+       * Verify that the bitcrusher does not alter samples beyond an acceptable
        * margin when parameterized to avoid information reduction.
        */
       function testBitcrusherInAudioContext() {
         // Sample Rate is held at a power of two to avoid sample interpolation
         // from the delay node. It is artificially low for easy inspection of
-        // buffers.   
+        // buffers.
         const sampleRate = 4096;
         let context = new OfflineAudioContext(2, sampleRate, sampleRate);
```
hoch: Using `sampleRate` twice doesn't look good. Please explain why or just use another `renderLength` variable.

audio-worklet/bitcrusher-test.html
```diff
@@ -155,17 +158,17 @@
...
       /**
-       * Verify that bitcrusher applies sample reduction on multiple channels of 
-       * input, and that sample reduction works across multiple buffers (which 
+       * Verify that bitcrusher applies sample reduction on multiple channels of
+       * input, and that sample reduction works across multiple buffers (which
        * requires bitcrusher memory.)
        */
       function testMultipleChannels() {
         const channelCount = 2;
         let context = new OfflineAudioContext(channelCount, 44100, 44100);
```
hoch: Use `sampleRate` and `renderLength` instead of hard-coded numbers.

audio-worklet/bitcrusher-test.html
```diff
@@ -155,17 +158,17 @@
...
         // Reduction should not be a factor of bufferSize for full coverage.
-        const reduction = 7; 
+        const reduction = 7;
         const bitDepth = 24;
-        const bufferSize = 512; 
+        const bufferSize = 512;
         // The two signals should differ to verify that the bitcrusher does not
         // place the results from one channel to another.
         let oscillator = new OscillatorNode(context, {frequency: 112});
```
hoch: Why 112?

audio-worklet/bitcrusher.html
```diff
@@ -13,46 +13,70 @@
...
+      <input type="radio" name="node" id="scriptProcessorButton" checked>
+          Script Processor<br>
+      <input type="radio" name="node" id="workletButton" value="worklet">
+          Audio Worklet<br>
+    </form>
+    <script>
+      // If the browser implements AudioWorklet, the demo enables the user to
+      // switch between audio worklet and script processor implementations of
+      // the bitcrusher.
+      const enableWorklet = AudioWorkletHelper.isAvailable();
```
hoch: This variable does not "enable" the worklet. Perhaps `workletIsAvailable` or `workletEnabled`?

audio-worklet/bitcrusher.html
```diff
@@ -13,46 +13,70 @@
...
+      if (enableWorklet) {
+        window.audioWorklet.addModule('src/bitcrusher-audio-worklet.js')
+            .then(() => {
+                const context = new AudioContext();
```
hoch: l.69-l.72 is identical to l.75-l.79 and I pointed that out before.

audio-worklet/lib/audio-worklet-helper.js
```diff
@@ -46,16 +45,15 @@ const AudioWorkletHelper = (() => {
...
                          'AudioWorklet is available. Start running demo...');
           demoFunc();
         } else {
           reportMessage_('error',
                          'The browser does not support AudioWorklet yet.');
         }
       });
-    }
-
+    },
```
hoch: We need a line break after this line.

audio-worklet/lib/audio-worklet-helper.js
```diff
@@ -46,16 +45,15 @@ const AudioWorkletHelper = (() => {
...
+    isAvailable: isAvailable_
```
hoch: Add a JSDoc comment for this exposed method.

audio-worklet/src/bitcrusher-audio-worklet.js
```diff
@@ -0,0 +1,77 @@
...
+      defaultValue: 1,
+      minValue: 1,
+      maxValue: 20
+    }];
+  }
+  
+  constructor(options) {
+    super(options);
+
+    // index_ and previousSample_ are member variables to handle block
```
hoch: Use the notation like `|index_|` for variables in the comment.

audio-worklet/src/bitcrusher-audio-worklet.js
```diff
@@ -0,0 +1,77 @@
...
+    // transitions. index_ refers to the current position in the input
+    // AudioBuffer, and previousSample_ refers to the corresponding sample.
+    this.index_ = 0;
+    this.previousSample_ = 0;
+  }
+
+  /**
+   * Bit crush upon receiving input AudioBuffer, applying lo-fi effects to the
```
hoch: - Bit crush > Bitcrush.
- Remove after "applying..." since we're not applying the effect onto the output buffer.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -15,25 +15,70 @@
  */
 class BitcrusherDemo {
   /**
+   * @class BitcrusherDemo
+   * @constructor
```
hoch: Pull the text next line up to here.

audio-worklet/src/bitcrusher-demo.js
```diff
@@ -15,25 +15,70 @@
...
-    this.masterGain_ = new GainNode(this.context_, {gain: (gain || 0.5)});
-    this.bitcrusher_ = new Bitcrusher(this.context_, {
-      channelCount: 1,
-      bitDepth: bitDepth || 24,
-      reduction: reduction || 1
-    });
+  constructor(context,  enableWorklet) {
+    this.context_ = context;
+    this.masterGain_ = new GainNode(this.context_, {gain: 0.5});
+    this.enableWorklet_ = enableWorklet;
```
hoch: This variable name is misleading to me.

rtoy: Haha. I thought the original `useWorklet` and suggested `enableWorklet` as a better name.

PR #96, opened 2017-08-02
Added testing utilities for noisegate

samhincks:
- audio-worklet/test/testharness.js is an exact copy of [testharness.js](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/resources/testharness.js?q=testharness.js&dr=C&l=2600) at 2017-08-02
- audio-worklet/test/audit.js is an exact copy of [audit.js](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/webaudio/resources/audit.js?q=audit.js&dr=C&l=39) at 2017-08-02
- audio-worklet/test/audit-util.js is an exact copy of [audit-util.js](https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/webaudio/resources/audit-util.js?dr=C) at 2017-08-02

### DIFF
--- a/audio-worklet/test/audit-util.js
+++ b/audio-worklet/test/audit-util.js
@@ -1,0 +1,348 @@
+// Copyright 2016 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+
+/**
+ * @fileOverview  This file includes legacy utility functions for the layout
+ *                test.
+ */
+
+// How many frames in a WebAudio render quantum.
+let RENDER_QUANTUM_FRAMES = 128;
+
+function writeString(s, a, offset) {
+  for (let i = 0; i < s.length; ++i) {
+    a[offset + i] = s.charCodeAt(i);
+  }
+}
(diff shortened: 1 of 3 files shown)


hoch:
Please say what you're trying to do with this PR in the description. Few bullet points would be nice.

hoch:
Are these exact copies from https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/webaudio/resources?

samhincks:
Yes these are exact copies. I just double checked with a diff that none of the files have changed in the Chromium repository since I last copied them.

hoch:
Then you can update the description with your last comment + a link to the source. Otherwise we'll have to look through the comment to figure out what this PR does.

rtoy:
Also, when you do the merge, github asks for a merge message.  This should be filled out more completely so that when I do a git log the message makes some sense.

(And yes, I am really terrible at doing this with github merge.  I usually write pretty good messages for my local commits, but the merge messages I enter are terrible.  I need to do a better job of this too.)

PR #97, opened 2017-08-04
Noise Gate Demo

samhincks:
This code enables a demo for the noise gate where the noise gate can be used to mute white noise during a pause in speech. There are 4 new files:

- noisegate.html  resembles the structure of bitcrusher.html
- src/noisegate-demo.js resembles the structure of src/bitcrusher-demo.js
- sound/white_noise.ogg is white noise 
- sound/speech.ogg  is a speech sample with a pause of the same duration as the noise file.

In addition, there is a slight modification to src/noisegate-script-processor.js. I changed the default threshold from 0 to -100. Zero does not make sense as a default threshold since that mutes all sound by default. In addition, I changed the logic supporting the default assignment of parameters. If the default value is nonzero, the current code does not allow for that default to be overriden to 
zero since (0 || |nonZeroValue|)  evaluates to a nonzero value.

The demo can be found at http://rawgit.com/GoogleChrome/web-audio-samples/fcca7efd31fe7a5e8f3b5a391150521b3ff55243/audio-worklet/noisegate.html


audio-worklet/noisegate.html
```diff
@@ -0,0 +1,87 @@
...
+      increasing the attack setting.
+    </p>
+    <div id="container"></div>
+    <form>
+      <input type="radio" name="node" id="scriptProcessorButton" checked>
+          Script Processor<br>
+      <input type="radio" name="node" id="workletButton" value="worklet">
+          Audio Worklet<br>
+    </form>
+    <form>
```
rtoy: Separating these forms/inputs would be useful to make it clear that these aren't related to the scriptprocessor/worklet choice would be good.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+    this.withoutNoisegateGain_.gain.setValueAtTime(0, t);
+
+    if (this.playing) {
+      this.attackSlider_.enable();
+      this.releaseSlider_.enable();
+      this.thresholdSlider_.enable();
+    }
+  }
+
+  withoutNoisegateSelected() {
```
hoch: What does this method do? The name is a bit weird.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+
+    this.attack_ = 0;
+    this.attackMax_ = 1;
+    this.attackMin_ = 0;
+    this.release_ = 0;
+    this.releaseMin_ = 0;
+    this.releaseMax_ = 1;
+    this.threshold_ = -100;
+    this.thresholdMin_ = -100;
+    this.thresholdMax_ = 0;
```
rtoy: Something is messed up.  I can set the threshold slider up to 100, which doesn't really make sense.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+        channelCount: 1,
+        attack: this.attack_,
+        release: this.release_,
+        threshold: this.threshold_
+    });
+
+    // There are two sound sources (speech and noise), which are mixed
+    // together. The user can set the gain of the noise, and also whether or
+    // not the mixed sound is processed by the noise gate.
+    this.speechGain_ = new GainNode(this.context_);
```
rtoy: This is really hard to follow.  A (very!) brief description of how the nodes are connected would be really helpful. I had to use the Audio Node Inspector extension to figure out what's connected.

rtoy: Also, why does the speech need it's own gain node?

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+    this.noiseGain_ = new GainNode(this.context_);
+    this.speechAndNoiseGain_ = new GainNode(this.context_);
+    this.withoutNoisegateGain_ = new GainNode(this.context_, {gain: 0});
+    this.noisegateGain_ = new GainNode(this.context_);
```
rtoy: Maybe call this `withNoisegateGain` to match `withouthNoisegateGain`? Probably also want to set explicitly the gain for this node to make the correspondence to `withoutNoisegateGain` clearer.

We're actually using these as a switch so maybe use better names?

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+        release: this.release_,
+        threshold: this.threshold_
+    });
+
+    // There are two sound sources (speech and noise), which are mixed
+    // together. The user can set the gain of the noise, and also whether or
+    // not the mixed sound is processed by the noise gate.
+    this.speechGain_ = new GainNode(this.context_);
+    this.noiseGain_ = new GainNode(this.context_);
+    this.speechAndNoiseGain_ = new GainNode(this.context_);
```
rtoy: You're basically using `speechAndNoiseGain` as a summing junction.  Give it a different name.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+    
+    this.noiseSource_.connect(this.noiseGain_);
+    this.speechSource_.connect(this.speechGain_);
+
+    this.speechSource_.onended = () => {
+      this.sourceButton_.enable();
+      this.attackSlider_.disable();
+      this.releaseSlider_.disable();
+      this.thresholdSlider_.disable();
+    }
```
hoch: Consider having a JS lint on you editor. That will complain about "missing semicolon" here.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+    this.speechAndNoiseGain_.connect(this.withoutNoisegateGain_)
+        .connect(this.masterGain_);
+    
+    this.speechAndNoiseGain_.connect(this.noisegateScriptProcessor_.input);
+    this.noisegateScriptProcessor_.output.connect(this.scriptProcessorGain_)
+        .connect(this.noisegateGain_).connect(this.masterGain_);
+
+    this.masterGain_.connect(this.context_.destination);
+
+    this.loadSound('sound/speech.ogg').then((sound) => {
```
rtoy: It would be nice if this speech segment were longer. I find it too short.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+    this.releaseSlider_.enable();
+    this.thresholdSlider_.enable();
+    this.playing = true;
+  }
+
+  /**
+   * Stop audio processing and configure UI elements.
+   */
+  stop() {
+    this.noiseSource_.stop();
```
hoch: Haven't thought about it deeply, but what happens if a new source starts before the previous source finishes? It seems like `this.fooSource_` is being overwritten, thus all the subsequent calls will be directed to the newly created source objects. WDYT?

audio-worklet/src/noisegate-script-processor.js
```diff
@@ -54,13 +54,21 @@ class NoiseGate {
...
     const bufferSize = options.bufferSize || 0;
-    this.threshold = options.threshold || 0;
     this.attack = options.attack || 0;
     this.release = options.release || 0;
     
     // The time constant of the filter has been set experimentally to balance
     // roughly delay for high frequency suppression.
-    const timeConstant = options.timeConstant || 0.0025;
+    let timeConstant;
+    if (options.timeConstant != null)
```
hoch: Use `undefined` and `!==`.

audio-worklet/src/noisegate-script-processor.js
```diff
@@ -54,13 +54,21 @@ class NoiseGate {
...
+      timeConstant = options.timeConstant;
+    else
+      timeConstant = 0.0025;
+
+    if (options.threshold != null)
```
hoch: Ditto.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+   *                  worklet should be used. If false, only the script
+   *                  processor will be used.
+   */
+  constructor(context, workletIsAvailable) {
+    this.context_ = context;
+    this.masterGain_ = new GainNode(this.context_, {gain: 0.5});
+    this.workletIsAvailable_ = workletIsAvailable;
+
+    this.attack_ = 0;
+    this.attackMax_ = 1;
```
rtoy: I thought we discussed (offline) that 1 sec might be too large. Is it not too large?

audio-worklet/src/noisegate-script-processor.js
hoch: Also you can rewrite this with the ternary operator.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+          type: 'range',
+          min: this.releaseMin_,
+          max: this.releaseMax_,
+          step: 0.01,
+          default: this.release_,
+          name: 'Release (s)'
+        });
+    this.releaseSlider_.disable();
+
+    this.gainSlider_ =
```
rtoy: `gainSlider` => `masterGainSlider`?  You have so many gain nodes it's hard to tell what this is.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+   * This is bound to an event listener by a ParamController.
+   * @param {Number} value The new threshold.
+   */
+  setThreshold(value) {
+    // TODO: When the param-controller update lands from the synthesizer branch,
+    // the stringToNumber conversion in this and subsequent methods should be
+    // removed.
+    const numericValue = parseInt(value);
+    if (numericValue < -100) throw 'The minimum threshold rate is -100.';
+    if (numericValue > 0) throw 'The maximum threshold rate is 0.';
```
rtoy: Are these things possible?  Doesn't the slider limit the values?

Probably didn't notice this in the noisegate design, but if there are limits on the threshold, shouldn't the noisegate implementation enforce those itself?

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+          default: this.release_,
+          name: 'Release (s)'
+        });
+    this.releaseSlider_.disable();
+
+    this.gainSlider_ =
+        new ParamController(containerId, this.setGain.bind(this), {
+          type: 'range',
+          min: 0,
+          max: 1,
```
rtoy: Why can't the master gain value have a larger maximum?  I think it should.

audio-worklet/src/noisegate-demo.js
```diff
@@ -0,0 +1,284 @@
...
+
+    this.speechSource_.onended = () => {
+      this.sourceButton_.enable();
+      this.attackSlider_.disable();
+      this.releaseSlider_.disable();
+      this.thresholdSlider_.disable();
+    }
+    
+    this.noiseSource_.loop = true;
+    this.speechSource_.loop = true;
```
rtoy: MIght as well do this in the ABSN constructors above.

audio-worklet/src/noisegate-script-processor.js
```diff
@@ -98,7 +106,7 @@ class NoiseGate {
   onaudioprocess_(event) {
     let inputBuffer = event.inputBuffer;
     let channel0 = inputBuffer.getChannelData(0);
-
+
```
rtoy: Remove.

hoch:
Can you suggest the right parameters for the demo? I tried a couple of config but none of them works as I expected.

audio-worklet/src/noisegate-demo.js
samhincks: I only put it there for symmetry, but I will take it out if that's the preferred style.

audio-worklet/src/noisegate-demo.js
samhincks: Perhaps you can show me how you use it on Monday. I tried it just now, and it seems to have a very low signal to noise ratio!

audio-worklet/src/noisegate-demo.js
samhincks: Yes, I'm not sure I see any problems here. The GUI ensures that stop is called before start can be called again. After the sources (`this.noiseSource_` and `this.speechSource_`) are stopped, then, by clicking start, these objects will be reinitialized and start again.

audio-worklet/src/noisegate-demo.js
samhincks: For the old demo (the one you heard on 08-04-2017), the audio sounded slightly better at around 250 ms. But for the new demo, higher attack and release don't buy you better sound since I make an effort in the recording to only pause in between sentences. So I will bring this back down to 0.1.

audio-worklet/src/noisegate-demo.js
samhincks: This was a problem with lib/param-controller.js and is fixed, as discussed in person.

audio-worklet/src/noisegate-demo.js
samhincks: I agree. I made it a bit longer and created a custom audio recording with no pauses within the sentence.

audio-worklet/src/noisegate-demo.js
hoch: This is a UI anti-pattern. The method is bound by the event listener, but it might throw an exception. Basically your app suddenly stops working when the slider touches the min/max value. The best thing you can do is to clamp it with a warning in the console.

audio-worklet/src/noisegate-demo.js
hoch: "a very low signal to noise ratio" - what do you mean by this?

There's nothing to it. If you install one via Package Control, it will just nag you when it finds nits like this in your code.

audio-worklet/src/noisegate-demo.js
hoch: Okay. Then I'll leave this up to you. I am fine this way.

audio-worklet/src/noisegate-demo.js
samhincks: Yes the slider limits these values for the threshold and also for attack and release. So if I take away the boundary check for the threshold, I should also take away the boundary check for attack and release.

The noise gate currently does not enforce threshold to be in the range -100 to 0. I will make this adjustment.

rtoy: In a different PR, please.  This PR is for the demo.

samhincks: Sure

audio-worklet/src/noisegate-script-processor.js
samhincks: I'm confused about what you want me to remove here. When I look at the file here, there is a blank line in between `let channel0 = inputBuffer.getChannelData(0); ` and the comment beneath it. Do you want me to remove the blank line? I think the blank line here is warranted since it's separating to logically distinct units of the code, and the next line is a comment.

rtoy: Sorry.  Remove the extraneous spaces on this line, leaving a single blank line without spaces in it.

samhincks: Got it. I just installed JSLint which is noticing this error. For some reason, this error wasn't caught with the TrailingSpaces package.

audio-worklet/noisegate.html
```diff
@@ -0,0 +1,81 @@
...
+      <input type="radio" name="node" id="withoutNoisegateButton"
+          value="withoutNoisegate">Without Noise Gate<br>
+    </form>
+    <script>
+      // TODO: Implement Audio Worklet support. For now, |workletIsAvailable|
+      // must be false.
+      const workletIsAvailable = false;
+      beginBitcrusherDemo();
+      
+      function beginBitcrusherDemo() {
```
hoch: Please fix the function name.

audio-worklet/noisegate.html
```diff
@@ -0,0 +1,81 @@
...
+      <input type="radio" name="node" id="noisegateButton" checked>
+          Noise Gate<br>
+      <input type="radio" name="node" id="withoutNoisegateButton"
+          value="withoutNoisegate">Without Noise Gate<br>
+    </form>
+    <script>
+      // TODO: Implement Audio Worklet support. For now, |workletIsAvailable|
+      // must be false.
+      const workletIsAvailable = false;
+      beginBitcrusherDemo();
```
hoch: This is a noisegate demo.

audio-worklet/noisegate.html
```diff
@@ -0,0 +1,81 @@
...
+    <p>
+      The volume of the noise (independent of the speech) can also be adjusted.
+      Try setting the noise to 0.25 and the threshold to -40db. With these
+      settings, the noise should be inaudible during the pause in speech.
+      There will be glitching between words, which can be minimized by
+      increasing the attack setting.
+    </p>
+    <div id="container"></div>
+    <form>
+      <input type="radio" name="node" id="noisegateButton" checked>
```
hoch: My previous comment was changing this to a toggle button of "active/bypass". Will you keep this as two radio buttons?

audio-worklet/noisegate.html
```diff
@@ -0,0 +1,81 @@
...
+    <script>
+      // TODO: Implement Audio Worklet support. For now, |workletIsAvailable|
+      // must be false.
+      const workletIsAvailable = false;
+      beginBitcrusherDemo();
+      
+      function beginBitcrusherDemo() {
+        const context = new AudioContext();
+        let demo = new NoisegateDemo(context, workletIsAvailable);
+        demo.initializeGUI('container', 'workletButton',
```
hoch: You don't have the `workletButton` in the DOM.

audio-worklet/src/noisegate-demo.js
hoch: When you tweak the preset, this might be useful: http://www.doctorproaudio.com/doctor/temas/dynamics-processors-noisegates_en.shtml

audio-worklet/src/noisegate-demo.js
hoch: Does the symmetry have any functional purpose? If not, that's redundant.

audio-worklet/src/noisegate-demo.js
hoch: Use the term `bypass` instead of `withoutSomething`. It's conventional and easy to understand.

audio-worklet/noisegate.html
samhincks: This has been changed on my end and will be a toggle in the next commit.

samhincks: I currently have this implemented as a checkbox. It seems like making this into a [toggle switch](https://www.w3schools.com/howto/howto_css_switch.asp) involves a lot of CSS, so I'm going to leave it as a checkbox for this PR.

hoch: https://github.com/GoogleChrome/omnitone/blob/master/examples/foa-renderer.html#L84

It's only 6~7 lines of JS code. CSS is not necessary at all.

samhincks:
The latest demo can be found [here](http://rawgit.com/GoogleChrome/web-audio-samples/101a06d34e4024d5a5a71a8f3c15b17dfd95226d/audio-worklet/noisegate.html). In the latest [commit](https://github.com/GoogleChrome/web-audio-samples/pull/97/commits/101a06d34e4024d5a5a71a8f3c15b17dfd95226d),
- I created a new, longer audio sample, where I try to speak quickly and take longer pauses. 
- With the noise gain at 1 and the threshold at -40db, the noise is inaudible during my pauses.
- A toggle button enables switching between active and bypass modes. 
- Other stylistic changes in response to comment.

audio-worklet/src/noisegate-demo.js
```diff
@@ -128,61 +129,63 @@ class NoisegateDemo {
...
-      this.attackSlider_.enable();
-      this.releaseSlider_.enable();
-      this.thresholdSlider_.enable();
-    }
-  }
+        this.attackSlider_.disable();
+        this.releaseSlider_.disable();
+        this.thresholdSlider_.disable();
+      }
+      else {
```
rtoy: Wrong formatting. It should be "} else {"

audio-worklet/src/noisegate-script-processor.js
```diff
@@ -56,19 +56,14 @@ class NoiseGate {
...
     // The time constant of the filter has been set experimentally to balance
     // roughly delay for high frequency suppression.
-    let timeConstant;
-    if (options.timeConstant != null)
-      timeConstant = options.timeConstant;
-    else
-      timeConstant = 0.0025;
+    let timeConstant = (typeof options.timeConstant === 'undefined') ?
+        0.0025 : options.timeConstant;
```
rtoy: What if I set `options.timeConstant = [1,2,3]`?

Don't think we need to fix this here in this PR, but if you're going to do checking here, maybe make it more robust?  Maybe check that options.timeConstant is a number (and finite?) and if it is, use it, otherwise use the default 0.0025?

audio-worklet/src/noisegate-demo.js
```diff
@@ -50,23 +50,25 @@ class NoisegateDemo {
...
-        .connect(this.noisegateGain_).connect(this.masterGain_);
+    this.noiseGain_.connect(this.speechAndNoiseSummingJunction_);
+    this.speechAndNoiseSummingJunction_.connect(
+        this.bypassNoisegateRoute_).connect(
+        this.masterGain_);
+
+    this.speechAndNoiseSummingJunction_.connect(
+        this.noisegateScriptProcessor_.input);
+    this.noisegateScriptProcessor_.output.connect(
+        this.scriptProcessorGain_).connect(
```
rtoy: What is the purpose of the `scriptProcessorGain`?

audio-worklet/src/noisegate-script-processor.js
samhincks: Sure, I will do this check as well as the threshold check in a later PR. Probably easiest to make this change the worklet integration PR. What do you think?

rtoy: Don't do this in this PR.

samhincks:
The new demo can be found at http://rawgit.com/GoogleChrome/web-audio-samples/372779777b3bc4b2c3c4f25674bbc79fbf916978/audio-worklet/noisegate.html

Issue #100, opened 2017-08-08
Poly-synth should support ADSR GUI controls


samhincks:
The poly-synth demo does not currently allow for the gain ADSR of a volume to change based on user input. In addition, the poly-synth demo does not support a filter ADSR. These features need to be implemented and controllable by the GUI.

PR #105, opened 2017-08-08
Implemented filter ADSR logic and volume and filter ADSR GUI controls.

samhincks:
This PR resolves [issue #100 ](https://github.com/GoogleChrome/web-audio-samples/issues/100). 
- poly-synth/index.html now has containers for filter and volume ADSR.
- poly-synth/src/poly-synth-demo.js now adds GUI elements for controlling filter and volume ADSR.
- poly-synth/src/poly-synth.js now has default values for filter and volume ADSR.
- poly-synth/src/poly-synth-voice.js now implements logic for filter ADSR.

The demo can be found [here](http://rawgit.com/GoogleChrome/web-audio-samples/55e6838d614479cfe39ea8b734005d4888079fb4/poly-synth/index.html). 

The parameters currently do not support nonlinear scaling. This is a desirable feature and entails modifying audio-worklet/lib/param-controllerer.js. This should be done in a separate PR, and is listed in issue #106

### DIFF
--- a/poly-synth/index.html
+++ b/poly-synth/index.html
@@ -20,15 +20,47 @@ limitations under the License.
     <script src="src/poly-synth-demo.js"></script>
     <script src="src/poly-synth.js"></script>
     <script src="src/poly-synth-voice.js"></script>
+    <style>
+      #container {
+        display: flex;
+        flex-direction: row;
+      }
+      .controls {
+        border-style: solid;
+        margin: 5px;
+      }
+    </style>
   </head>
   <body>
+    <h1>Polyphonic Subtractive Synthesizer</h1>
     <p>
-     The key A on your computer keyboard maps onto the note C, and the letters
-     to the right of A (SDFGHJ) map onto the notes above C in a C major scale
-     (DEFGAB). The letters WETYU map onto the black keys of a piano.
+      The key A on your computer keyboard maps onto the note C, and the letters
+      to the right of A (SDFGHJ) map onto the notes above C in a C major scale
+      (DEFGAB). The letters WETYU map onto the black keys of a piano.
+    </p>
+    <p>
+      The synthesizer constructs voices according to the volume and filter ADSR
+      settings. The volume ADSR modulates the gain of the signal over time, and
+      the filter ADSR modulates the detune of the cutoff filter, meaning the
+      cutoff specifies the filter's starting point and it will detune over
+      time depending on the ADSR settings. A detune amount of 1 means the filter
+      will deviate by at most 2400 cents from the filter cutoff value. For both
+      the volume and filter ADSR,
+      <ul>
+        <li>Attack refers to the number of seconds until the maximum level.</li>
+        <li>Decay refers to the number of seconds between the attack level
+            and the sustain level.</li>
+        <li>Sustain refers to the steady value as the note is pressed as
+            a fraction of the maximum level.</li>
+        <li>Release refers to the number of seconds from releasing the note
+            until a zero value.</li>
+      </ul>
     </p>
     <div id="keyboard"></div>
-    <div id="container"></div>
+    <div id="container">
+      <div id="gainADSR" class="controls"><h2>Volume</h2></div>
+      <div id="filterADSR" class="controls"><h2>Filter</h2></div>
+    </div>
     <p>
       Credit to Stuart Memo for the
       <a href="https://github.com/stuartmemo/qwerty-hancock">keyboard</a>.
@@ -36,7 +68,7 @@ limitations under the License.
     <script>
       const context = new AudioContext();
       let polySynthDemo = new PolySynthDemo(context);
-      polySynthDemo.initializeGUI('container');
+      polySynthDemo.initializeGUI('gainADSR', 'filterADSR');
     </script>
   </body>
 </html>
--- a/poly-synth/index.html
+++ b/poly-synth/index.html
@@ -21,12 +21,11 @@ limitations under the License.
     <script src="src/poly-synth.js"></script>
     <script src="src/poly-synth-voice.js"></script>
     <style>
-      #container {
+      #param-container {
         display: flex;
         flex-direction: row;
       }
       .controls {
-        border-style: solid;
         margin: 5px;
       }
     </style>
@@ -39,13 +38,9 @@ limitations under the License.
       (DEFGAB). The letters WETYU map onto the black keys of a piano.
     </p>
     <p>
-      The synthesizer constructs voices according to the volume and filter ADSR
-      settings. The volume ADSR modulates the gain of the signal over time, and
-      the filter ADSR modulates the detune of the cutoff filter, meaning the
-      cutoff specifies the filter's starting point and it will detune over
-      time depending on the ADSR settings. A detune amount of 1 means the filter
-      will deviate by at most 2400 cents from the filter cutoff value. For both
-      the volume and filter ADSR,
+      The synthesizer constructs voices according to the amplifier and filter
+      ADSR settings. The amplifier ADSR modulates the gain of the signal over
+      time, and the filter ADSR modulates the detune of the cutoff filter.
       <ul>
         <li>Attack refers to the number of seconds until the maximum level.</li>
         <li>Decay refers to the number of seconds between the attack level
@@ -56,10 +51,12 @@ limitations under the License.
             until a zero value.</li>
       </ul>
     </p>
-    <div id="keyboard"></div>
-    <div id="container">
-      <div id="gainADSR" class="controls"><h2>Volume</h2></div>
-      <div id="filterADSR" class="controls"><h2>Filter</h2></div>
+    <div id="synth-container">
+      <div id="keyboard"></div>
+      <div id="param-container">
+        <div id="amp-envelope" class="controls"><h2>Amplifier</h2></div>
+        <div id="filter-envelope" class="controls"><h2>Filter</h2></div>
+      </div>
     </div>
     <p>
       Credit to Stuart Memo for the
@@ -68,7 +65,7 @@ limitations under the License.
     <script>
       const context = new AudioContext();
       let polySynthDemo = new PolySynthDemo(context);
-      polySynthDemo.initializeGUI('gainADSR', 'filterADSR');
+      polySynthDemo.initializeGUI('amp-envelope', 'filter-envelope');
     </script>
   </body>
 </html>
--- a/poly-synth/src/poly-synth-demo.js
+++ b/poly-synth/src/poly-synth-demo.js
@@ -42,24 +42,14 @@ class PolySynthDemo {
 
   /**
    * Initialize GUI components.
-   * @param {String} containerId The ID of the container element (e.g. <div>).
+   * @param {String} gainADSRId The ID of the container for gain ADSR elements
+   *                            (e.g. <div>).
+   * @param {String} filterADSRId The ID of the container for filter
+   *                              ADSR elements.
    */
-  initializeGUI(containerId) {
-    let lowPassSlider_ = new ParamController(
-        containerId, this.polySynth_.setParameter.bind(this.polySynth_), {
-          name: 'Cutoff',
-          id: 'cutoff',
-          type: 'range',
-          min: this.polySynth_.lowPassMinCutoff,
-          max: this.polySynth_.lowPassMaxCutoff,
-          step: 1,
-          default: this.polySynth_.lowPassMaxCutoff,
-        });
-
-    // The maximum gain is set to be larger than 1 to allow for highly filtered
-    // sounds to be audible. It is up to the user to avoid distortion.
+  initializeGUI(gainADSRId, filterADSRId) {
     let masterGainSlider_ =
-        new ParamController(containerId, this.setGain.bind(this), {
+        new ParamController(gainADSRId, this.setGain.bind(this), {
           name: 'Volume',
           id: 'masterGain',
           type: 'range',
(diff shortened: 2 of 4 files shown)


poly-synth/index.html
```diff
@@ -20,23 +20,55 @@
     <script src="src/poly-synth-demo.js"></script>
     <script src="src/poly-synth.js"></script>
     <script src="src/poly-synth-voice.js"></script>
+    <style>
+      #container {
+        display: flex;
+        flex-direction: row;
+      }
+      .controls {
+        border-style: solid;
```
hoch: Is this really necessary?

poly-synth/index.html
```diff
@@ -20,23 +20,55 @@
...
+    </p>
+    <p>
+      The synthesizer constructs voices according to the volume and filter ADSR
+      settings. The volume ADSR modulates the gain of the signal over time, and
+      the filter ADSR modulates the detune of the cutoff filter, meaning the
+      cutoff specifies the filter's starting point and it will detune over
+      time depending on the ADSR settings. A detune amount of 1 means the filter
+      will deviate by at most 2400 cents from the filter cutoff value. For both
+      the volume and filter ADSR,
+      <ul>
```
hoch: I find this description is too verbose and confusing. Just having 3~4 bullet points about how to use the synth would be good enough. You don't need to explain how synthesizer works here.

poly-synth/index.html
```diff
@@ -20,23 +20,55 @@
...
+            and the sustain level.</li>
+        <li>Sustain refers to the steady value as the note is pressed as
+            a fraction of the maximum level.</li>
+        <li>Release refers to the number of seconds from releasing the note
+            until a zero value.</li>
+      </ul>
     </p>
     <div id="keyboard"></div>
-    <div id="container"></div>
+    <div id="container">
```
hoch: Perhaps `id = "synth-container"`? Then move the keyboard above inside of this container.

poly-synth/index.html
```diff
@@ -20,23 +20,55 @@
...
+      <div id="gainADSR" class="controls"><h2>Volume</h2></div>
```
hoch: Why `h2`? Also perhaps `amp-envelope` is the better name.

poly-synth/index.html
```diff
@@ -20,23 +20,55 @@
...
+      <div id="filterADSR" class="controls"><h2>Filter</h2></div>
```
hoch: Ditto: `filter-envelope`

poly-synth/index.html
```diff
@@ -20,23 +20,55 @@
...
+        flex-direction: row;
+      }
+      .controls {
+        border-style: solid;
+        margin: 5px;
+      }
+    </style>
   </head>
   <body>
+    <h1>Polyphonic Subtractive Synthesizer</h1>
```
hoch: I hope you can run `html-tidy` on this file.

poly-synth/src/poly-synth-demo.js
```diff
@@ -76,7 +176,7 @@ class PolySynthDemo {
    * @param {Number} value The new gain.
    */
   setGain(value) {
-    this.masterGain_.gain.value = parseFloat(value);
+    this.masterGain_.gain.value = value;
```
rtoy: Why remove the `parseFloat`? Was this a mistake in the previous code that we overlooked?

poly-synth/src/poly-synth-voice.js
```diff
@@ -58,12 +58,35 @@ class PolySynthVoice {
   start() {
     // Ramp to full amplitude in attack (s) and to sustain in decay (s).
     const t = this.context_.currentTime;
-    const timeToFullAmplitude = t + this.parameters_.attack;
-    const timeToSustain = timeToFullAmplitude + this.parameters_.decay;
+    const timeToFullAmplitude = t + this.parameters_.gainAttack;
+
+    // |timeToGainSustain| must be later than |timeToFullAmplitude|.
+    const timeToGainSustain =
+        timeToFullAmplitude + this.parameters_.gainDecay + 0.001;
```
rtoy: Why 0.001?

poly-synth/src/poly-synth-voice.js
```diff
@@ -58,12 +58,35 @@ class PolySynthVoice {
...
-        this.parameters_.sustain, timeToSustain);
+        this.parameters_.gainSustain, timeToGainSustain);
+
+    // The detune of the filter reaches its peak amount specified by
+    // |filterDetuneAmount| (where 1 corresponds to 2400 cents detuning) in
+    // |filterAttack| seconds. It then decays to a fraction of that amount as
+    // specified by |filterSustain| in |filterDecay| seconds.
+    const standardFilterDetuneInCents = 2400;
+    const amountOfFilterDetuneInCents =
+        standardFilterDetuneInCents * this.parameters_.filterDetuneAmount;
```
rtoy: Why not have `filterDetuneAmount` already be in units of cents?

poly-synth/src/poly-synth-voice.js
```diff
@@ -58,12 +58,35 @@ class PolySynthVoice {
...
+    const amountOfSustainDetuneInCents =
+        standardFilterDetuneInCents * this.parameters_.filterSustain;
+
+    const timeToFullDetune = t + this.parameters_.filterAttack;
+    const timeToDetuneSustain
+        = timeToFullDetune + this.parameters_.filterDecay + 0.001;
```
rtoy: Again, why 0.001?

poly-synth/src/poly-synth.js
```diff
@@ -32,15 +32,38 @@ class PolySynth {
     this.releasedVoices_ = [];
 
     // The maximum and minimum cutoffs are experimentally determined.
-    this.lowPassMaxCutoff = 16000;
-    this.lowPassMinCutoff = 60;
+    this.maxCutoff = 16000;
```
rtoy: Why the name change?

poly-synth/src/poly-synth.js
```diff
@@ -32,15 +32,38 @@ class PolySynth {
...
+    // well as the ADSR for the filter. These values are experimentally
+    // determined.
+    this.maxAttack = 10;
+    this.minAttack = 0;
+    this.minSustain = 0.1;
+    this.maxSustain = 1;
+    this.minDecay = 0;
+    this.maxDecay = 10;
+    this.minRelease = 0;
+    this.maxRelease = 10;
```
rtoy: What makes these values correct?  I don't think there's anything for you to do here, but I'm just curious.

poly-synth/src/poly-synth.js
```diff
@@ -32,15 +32,38 @@ class PolySynth {
...
+    this.maxDetuneAmount = 5;
+
+    // The initial values for the parameters are experimentally determined.
     this.parameters_ = {
-      cutoff: this.lowPassMaxCutoff,
-      attack: 0,
-      decay: 0,
-      sustain: 0.1,
-      release: 0
+      filterCutoff: 440,
```
rtoy: Why a default cutoff of 440 Hz?  Doesn't that mean that if I play the highest note on the keyboard that I hear almost nothing?

rtoy:
After playing around with the demo, I have some comments (based on my lack of any recent experience with real synthesizers):

- There's a click at the start and end of every key I press.  Shouldn't the gain ADSR defaults be set to something so that the click doesn't happen?
- The gain ADSR sliders have a range of up to 10 s.  Is that realistic?
- The Volume section might have a different name to reflect that it's per voice gain ADSR?  Not sure, though.
- The filter cutoff slider range is huge and very hard to control.  Seems like this would be a fairly simple change to the param controller stuff.  Add a function to the param controller that takes the slider value and a function to convert the slider value to a printed value.  This would also be the output value of the slider.

And there's a bug somewhere where you're not releasing notes. I picked a huge release and moved the mouse over all the keys randomly, generating a huge number of notes.  Somehow, I now have a note or two sustained apparently indefinitely.  Along with some clicks.

It's a known issue in Chrome that biquads liver forever.  This probably isn't helping with the CPU usage/clicks.  A work around is to disconnect the biquad shortly after the oscillator stops.

poly-synth/index.html
samhincks: I think the border makes it clearer which ADSR parameters refer to the filter and which refer to the amplifier

poly-synth/src/poly-synth.js
hoch: Also note that the member variables below do not have a trailing underscore.

poly-synth/index.html
samhincks: I couldn't find a sublime text package called HTML tidy but I noticed that [SublimeLinter](http://www.sublimelinter.com/en/latest/) had an HTML tidy component. I ran that and it gave warnings about these lines 
```
<script src="third-party/qwerty-hancock.min.js"></script>
    <script src="../audio-worklet/lib/param-controller.js"></script>
    <script src="src/poly-synth-demo.js"></script>
    <script src="src/poly-synth.js"></script>
    <script src="src/poly-synth-voice.js"></script>
``` saying inserting "type" attribute. It wasn't complaing about the line you reference. What is wrong with it? And am I using the correct html-tidy program?

File: poly-synth/index.html (hunk 5 above)
Comment (samhincks) [inline poly-synth/index.html]: In JS I use the camelCase variable name style. Is the correct style to use the hyphen-based variable naming in HTML?

File: poly-synth/src/poly-synth-demo.js (hunk 7 above)
Comment (samhincks) [inline poly-synth/src/poly-synth-demo.js]: Using `parseFloat` was correct so long as lib/param-controller.js did not to string to number conversion. I think that change to lib/param-controller.js got landed before the poly-synth, so, yes this was a mistake that we overlooked.

File: poly-synth/src/poly-synth-voice.js (hunk 8 above)
Comment (samhincks) [inline poly-synth/src/poly-synth-voice.js]: The reason I added a small constant to` timeToFullAmplitude + this.parameters_.gainDecay` is that timeToGainSustain cannot be the same as timeToFullAmplitude since the two linearRampToValueAtTime commands will conflict. Thinking some more about it, the best solution seems to me to be to set the minimum decay time to a small positive value.

File: poly-synth/src/poly-synth-voice.js (hunk 9 above)
Comment (samhincks) [inline poly-synth/src/poly-synth-voice.js]: The convention according to @hoch is to let `filterDetuneAmount` be a multiplier of a standard detune amount (2400). So when `filterDetuneAmount` is 0, there is no detuning and when it is its maximum value 5, the detuning is 1200 (5 *2400).

File: poly-synth/src/poly-synth-voice.js (hunk 10 above)
Comment (samhincks) [inline poly-synth/src/poly-synth-voice.js]: See response above.

File: poly-synth/src/poly-synth-demo.js (hunk 7 above)
Comment (rtoy) [inline poly-synth/src/poly-synth-demo.js]: Ok.

File: poly-synth/src/poly-synth-voice.js (hunk 9 above)
Comment (rtoy) [inline poly-synth/src/poly-synth-voice.js]: Ok.

File: poly-synth/src/poly-synth.js (hunk 11 above)
Comment (samhincks) [inline poly-synth/src/poly-synth.js]: The synthesizers I've looked at don't distinguish if its using a low pass or a high pass filter unless it exposes both, so partly I didn't think it was necessary and partly, the code reads better with this shorter variable name.

Comment (rtoy) [inline poly-synth/src/poly-synth.js]: Be that as it may, churning names in the middle of a review when no reviewer asked for such a change just makes the review take longer since there are now more things to review that didn't need to be reviewed.

File: poly-synth/src/poly-synth.js (hunk 12 above)
Comment (samhincks) [inline poly-synth/src/poly-synth.js]: These are the values I've seen in Logic Pro, but Logic Pro enables nonlinear adjustment for these parameters. I can see argument for making them shorter if we don't support logarithmic control. 

At smaller values for (attack, release, and decay), these properties allow different timbral qualities, which are interesting for melodies and rhythms etc. At larger values, these controls allow the musician to create more long term movements of a song, e.g. a long filter sweep drowning out the other parts of a song, bringing the song into a different chapter.

File: poly-synth/src/poly-synth.js (hunk 13 above)
Comment (samhincks) [inline poly-synth/src/poly-synth.js]: That's a good point. I haven't implemented higher octaves yet, so we currently can't play higher notes. 
I liked the sound most at this cutoff value, which is why I chose it. I think the synthesizer should sound good in it's default setting. But I agree we will need to reconsider this default value if we implement more octaves.

File: poly-synth/src/poly-synth.js (hunk 11 above)
Comment (samhincks) [inline poly-synth/src/poly-synth.js]: Sure, I'll try not to do that again.

File: poly-synth/src/poly-synth.js [hunk 14]
```diff
@@ -52,13 +55,17 @@ class PolySynth {
     this.minDetuneAmount = 0;
     this.maxDetuneAmount = 5;
 
+    this.minQ = 0;
+    this.maxQ = 100;
```
rtoy: The Q parameter for a lowpass filter is in dB.  Read the spec and choose different values. Or play around with http://googlechrome.github.io/web-audio-samples/samples/audio/mag-phase.html to see what makes sense. (There's a bug in that it doesn't know that the Q for lowpass/highpass filters is in dB.  I need to fix that.)

samhincks:
In the latest [commit](https://github.com/GoogleChrome/web-audio-samples/pull/105/commits/23d0f8ddc5b5de02c7df0339d0167a6de6c51bf9), 
- I added more JSDoc to getParameters(). This was overlooked in the last review (the old documentation didn't document the filter ADSR)
- I tweaked default parameter variables. Decay can never be zero since the linear ramp to sustain level and max level will conflict with each other when the timing is identical. The default release is >0 to avoid glitching.
- I added Q
- Other modifications based on review.

Ray noticed a bug in the synthesizer, which I investigated, and the simplest way to reproduce it is to click a note with the keyboard and then to click that note again with the mouse while it is pressed. The note will continue forever. I will file an issue. The [demo](http://rawgit.com/GoogleChrome/web-audio-samples/23d0f8ddc5b5de02c7df0339d0167a6de6c51bf9/poly-synth/index.html) can be found here.

poly-synth/src/poly-synth.js
```diff
@@ -52,13 +55,17 @@ class PolySynth {
...
     this.parameters_ = {
-      filterCutoff: 440,
-      gainAttack: this.minAttack,
+      gainAttack: 0,
       gainDecay: this.minDecay,
       gainSustain: this.minSustain,
-      gainRelease: this.minRelease,
+      gainRelease: 0.1,
+      filterCutoff: 440,
+      filterQ: this.minQ,
```
rtoy: Why minQ?

poly-synth/src/poly-synth.js
```diff
@@ -74,12 +81,24 @@ class PolySynth {
...
+   *                                           filter.
+   * @returns {Number} parameters.filterQ The peak of the response at the cutoff
+   *                                      frequency.
+   * @returns {Number} parameters.filterAttack Seconds until full detune.
+   * @returns {Number} parameters.filterDecay Seconds between full detune and
+   *                                          sustain detune.
+   * @returns {Number} parameters.filterSustain The steady detune level as a
+   *                                            note is pressed.
+   * @returns {Number} parameters.filterRelease Seconds betweeen the release of
+   *                                            a note and zero detune.
```
rtoy: Probably want to say the audio params are k-rate not a-rate somewhere around here.

poly-synth/src/poly-synth.js
```diff
@@ -32,15 +32,45 @@ class PolySynth {
...
+    // The initial values for the parameters are experimentally determined.
     this.parameters_ = {
-      cutoff: this.lowPassMaxCutoff,
-      attack: 0,
-      decay: 0,
-      sustain: 0.1,
-      release: 0
+      gainAttack: 0,
+      gainDecay: this.minDecay,
+      gainSustain: this.minSustain,
```
rtoy: Is minSustain (0.1?) really the right value for the default?  Pretty drastic change in volume from attack to sustain.

poly-synth/src/poly-synth.js
samhincks: Q has a pretty profound effect on the sound, and adds a lot of character. I think it's better to leave it "off" initially and let the user increase it if they so desire.

poly-synth/src/poly-synth.js
samhincks: The advantage with a lower value is that the sounds are more emphasized when you hit the note, giving the instrument a sharper edge. However, I think you're right that you don't want the synthesizer to default to large differences in volume over the course of the note, so I will increase this value to 0.5 as a compromise.

samhincks:
In the latest [commit](https://github.com/GoogleChrome/web-audio-samples/pull/105/commits/55589535a1f17639946dbb3e9349976b705e434f), 
- Modifications based on review
- I have not changed sustain to sustain volume in the GUI since that is not the convention in other DAWs I have seen. This information is "subtly" conveyed by the other parameters having a seconds unit, e.g (Attack (s)). 

The latest demo is now [here](http://rawgit.com/GoogleChrome/web-audio-samples/55589535a1f17639946dbb3e9349976b705e434f/poly-synth/index.html).

Issue #102, opened 2017-08-08
Audio worklet counterpart to noisegate needs to be implemented


samhincks:
Currently, there is only a script processor implementation of the noise gate. It needs to be implemented as an audio worklet as well.

PR #107, opened 2017-08-09
Added a worklet implementation of the noisegate.

samhincks:
This PR has a single file implementing a worklet counterpart to audio-worklet/src/noisegate-script-processor.js. It resolves issue #102.

### DIFF
--- a/audio-worklet/src/noisegate-audio-worklet.js
+++ b/audio-worklet/src/noisegate-audio-worklet.js
@@ -77,6 +77,8 @@ registerProcessor('noisegate-audio-worklet',
     this.previousWeight_ = 1.0;
     this.envelope_ = new Float32Array(128);
     this.weights_ = new Float32Array(128);
+
+    // TODO: Use getContextInfo() to get sample rate.
     this.sampleRate = 44100;
   }
 
@@ -170,8 +172,8 @@ registerProcessor('noisegate-audio-worklet',
     // a^2 / 2. Sine waves are therefore scaled back to the original
     // amplitude, but other waveforms or constant sources can only be
     // approximated.
-    const scaledEnvelopeValue;
-    const weight;
+    let scaledEnvelopeValue;
+    let weight;
 
     // Compute an array of weights between 0 and 1 which will be multiplied with
     // the channel depending on if the noise gate is open, attacking, releasing,
--- a/audio-worklet/src/noisegate-audio-worklet.js
+++ b/audio-worklet/src/noisegate-audio-worklet.js
@@ -78,7 +78,7 @@ registerProcessor('noisegate-audio-worklet',
     this.envelope_ = new Float32Array(128);
     this.weights_ = new Float32Array(128);
 
-    // TODO: Use getContextInfo() to get sample rate.
+    // TODO (issue #111): Use getContextInfo() to get sample rate.
     this.sampleRate = 44100;
   }
 
@@ -86,7 +86,7 @@ registerProcessor('noisegate-audio-worklet',
    * Control the dynamic range of input based on specified threshold.
    * @param {AudioBuffer} input Input audio data
    * @param {AudioBuffer} output Output audio data
-   * @param {Object} parameters
+   * @param {Object} parameters K-rate audio params.
    * @param {Number} parameters.attack Seconds for gate to fully close.
    * @param {Number} parameters.release Seconds for gate to fully open.
    * @param {Number} parameters.timeConstant Seconds for envelope follower's
--- a/audio-worklet/src/noisegate-audio-worklet.js
+++ b/audio-worklet/src/noisegate-audio-worklet.js
@@ -1,0 +1,194 @@
+/**
+ * Copyright 2017 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @class NoiseGate
+ * @extends AudioWorkletProcessor
+ * @description A noise gate allows audio signals to pass only when the
+ * registered volume is above a specified threshold.
+ */
+registerProcessor('noisegate-audio-worklet',
+                  class NoiseGateAudioWorklet extends AudioWorkletProcessor {
+  
+  static get parameterDescriptors() {
+    return [{
+      name: 'attack',
+      defaultValue: 0,
+      minValue: 0,
+      maxValue: 10
+    }, {
+      name: 'release',
+      defaultValue: 0,
+      minValue: 0,
+      maxValue: 10
+    }, {
+      name: 'threshold',
+      defaultValue: -40,
+      minValue: -100,
+      maxValue: 0
+    }, {
+      name: 'timeConstant',
+      defaultValue: 0.0025,
+      minValue: 0,
+      maxValue: 1
+    }];
+  }
+
+  constructor() {
+    super();
+
+    // The previous envelope level, a float representing signal amplitude.
+    this.previousLevel_ = 0;
+
+    // The last weight (between 0 and 1) assigned, where 1 means the gate
+    // is open and 0 means it is closed and the sample in the output buffer is
+    // muted. When attacking, the weight will linearly decrease from 1 to 0, and
+    // when releasing the weight linearly increase from 0 to 1.
+    this.previousWeight_ = 1.0;
+    this.envelope_ = new Float32Array(128);
+    this.weights_ = new Float32Array(128);
+    this.sampleRate = 44100;
+  }
+
+  /**
+   * Control the dynamic range of input based on specified threshold.
+   * @param {AudioBuffer} input Input audio data
+   * @param {AudioBuffer} output Output audio data
+   * @param {Object} parameters
+   * @param {Number} parameters.attack Seconds for gate to fully close.
+   * @param {Number} parameters.release Seconds for gate to fully open.
+   * @param {Number} parameters.timeConstant Seconds for envelope follower's
+   *                                         smoothing filter delay.
+   * @param {Number} parameters.threshold Decibel level beneath which sound is
+   *                                      muted.
+   */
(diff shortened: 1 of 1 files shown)


audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,194 @@
...
+      name: 'threshold',
+      defaultValue: -40,
+      minValue: -100,
+      maxValue: 0
+    }, {
+      name: 'timeConstant',
+      defaultValue: 0.0025,
+      minValue: 0,
+      maxValue: 1
+    }];
```
rtoy: Add comments on how you picked these default values and the ranges.

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,194 @@
...
+    this.previousLevel_ = 0;
+
+    // The last weight (between 0 and 1) assigned, where 1 means the gate
+    // is open and 0 means it is closed and the sample in the output buffer is
+    // muted. When attacking, the weight will linearly decrease from 1 to 0, and
+    // when releasing the weight linearly increase from 0 to 1.
+    this.previousWeight_ = 1.0;
+    this.envelope_ = new Float32Array(128);
+    this.weights_ = new Float32Array(128);
+    this.sampleRate = 44100;
```
rtoy: Why 44100?

And why not trailing underscore?

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,194 @@
...
+   *                                         smoothing filter delay.
+   * @param {Number} parameters.threshold Decibel level beneath which sound is
+   *                                      muted.
+   */
+  process(input, output, parameters) {
+    // Alpha controls a tradeoff between the smoothness of the
+    // envelope and its delay, with a higher value giving more smoothness at
+    // the expense of delay and vice versa.
+    this.alpha_ = this.getAlphaFromTimeConstant_(
+        parameters.timeConstant[0], 44100);
```
rtoy: You should at least use `this.sampleRate` instead of 44100.

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,194 @@
...
+
+    this.attack = parameters.attack[0];
```
hoch: Please add a comment; these are 'k-rate' params if you use them this way.

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,194 @@
...
+
+    // Compute an array of weights between 0 and 1 which will be multiplied with
+    // the channel depending on if the noise gate is open, attacking, releasing,
+    // or closed.
+    for (let i = 0; i < envelope.length; i++) {
+      // For sine waves, the envelope eventually reaches an average power of
+      // a^2 / 2. Sine waves are therefore scaled back to the original
+      // amplitude, but other waveforms or constant sources can only be
+      // approximated.
+      const scaledEnvelopeValue
```
hoch: I would pull the local variables here outside of `for` loop. They alway are overwritten for each iteration, so no need to recreate the variables every time. (i.e. less garbage)

audio-worklet/src/noisegate-audio-worklet.js
hoch: `AudioWorkletProcessor` has `getContextInfo()` but it's not available in the custom build.

audio-worklet/src/noisegate-audio-worklet.js
hoch: Because the sample rate is provided by the global scope.

samhincks: So I should leave it as it is? This is sort of an edge case. I'm hardcoding the sample rate to be 44,100 but I want to match a variable in the global scope. I can file an issue to change this once getContextInfo() is implemented.

audio-worklet/src/noisegate-audio-worklet.js
samhincks: I should say that the a-rate audio params are used as k-rate audio params since the current build does not support k-rate audio params?

audio-worklet/src/noisegate-audio-worklet.js
samhincks: Sure... I'm just curios. Are we more concerned about memory efficiency in the worklet version as opposed to the script processor version?

audio-worklet/src/noisegate-audio-worklet.js
rtoy: Or, maybe you can just actually make them a-rate?  But not in this PR.

samhincks:
In the latest [commit](https://github.com/GoogleChrome/web-audio-samples/pull/107/commits/30e030378289cc62db168a192dec149a8a92d708), 
- The default values and ranges for audio params have been changed. And I added comments for these.
- Other minor changes based on review.

audio-worklet/src/noisegate-audio-worklet.js
rtoy: Add a TODO comment here to say it should use getContextInfo() to get the sample rate.

audio-worklet/src/noisegate-audio-worklet.js
rtoy: This should also be mentioned in the JSDoc for this node.

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -148,22 +166,24 @@ registerProcessor('noisegate-audio-worklet',
       releaseSteps = Math.ceil(this.sampleRate * this.release);
       releaseGainPerStep = 1 / releaseSteps;
     }
+    // For sine waves, the envelope eventually reaches an average power of
+    // a^2 / 2. Sine waves are therefore scaled back to the original
+    // amplitude, but other waveforms or constant sources can only be
+    // approximated.
+    const scaledEnvelopeValue;
+    const weight;
```
rtoy: How does this work?  These are const, but aren't initialized. And since they're const, they can't be changed.

samhincks:
In the latest [commit](https://github.com/GoogleChrome/web-audio-samples/pull/107/commits/c8ad8d4412a9088fb244121e52504226f1e65652), 
- Changed const to let (the code was submitted with a syntax error previously)
- Added a TODO comment.

audio-worklet/src/noisegate-audio-worklet.js
rtoy: If you filed an issue on this, replace "TODO" with "TODO(#n)" where n is the issue number.  Or "TODO(url-to-issue)".

If you haven't filed an issue, file it, and do the above.

audio-worklet/src/noisegate-audio-worklet.js
rtoy: Yes.  In the script processor if you generate garbage, you'll end up messing up the main thread and possibly the audio thread if GC takes too long.

In a worklet, if you generate garbage, you're pretty much guaranteed to glitch audio because GC on Chrome can take 10ms or more.

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,216 @@
...
+registerProcessor('noisegate-audio-worklet',
+                  class NoiseGateAudioWorklet extends AudioWorkletProcessor {
+  
+  static get parameterDescriptors() {
+    return [
+    // An upper bound of 100ms for attack and release is sufficiently high
+    // to enable smooth transitions between sound and silence.
+    // The default value of 50ms has been set experimentally to minimize
+    // glitches in the demo.
+    {
```
hoch: Have you applied clang-format on this?

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,216 @@
...
+      minValue: 0,
+      maxValue: 0.1
+    },
+    {
+      name: 'release',
+      defaultValue: 0.05,
+      minValue: 0,
+      maxValue: 0.1
+    },
+    // The maximum threshold is 0 since that is the defined maximum in dBFS.
```
hoch: I still find the comments are too verbose, but it's up to you.

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,216 @@
...
+   *                                      muted.
+   */
+  process(input, output, parameters) {
+    // Alpha controls a tradeoff between the smoothness of the
+    // envelope and its delay, with a higher value giving more smoothness at
+    // the expense of delay and vice versa.
+    this.alpha_ = this.getAlphaFromTimeConstant_(
+        parameters.timeConstant[0], this.sampleRate);
+
+    // The a-rate audio-params are used as k-rate audio-params.
```
hoch: Perhaps

`// k-rate: use the first element of parameter data to process each render quantum.`

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,216 @@
...
+   */
+  computeWeights_(envelope) {
+    // When attack or release are 0, the weight changes between 0 and 1
+    // in one step.
+    let attackSteps = 1;
+    let releaseSteps = 1;
+    let attackLossPerStep = 1;
+    let releaseGainPerStep = 1;
+
+    // TODO: Replace this weights-based approach for enabling attack/release
```
hoch: I think you should move this comment to l.149.

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,216 @@
...
+    // or release time parameters.
+    if (this.attack > 0) {
+      attackSteps = Math.ceil(this.sampleRate * this.attack);
+      attackLossPerStep = 1 / attackSteps;
+    }
+    if (this.release > 0) {
+      releaseSteps = Math.ceil(this.sampleRate * this.release);
+      releaseGainPerStep = 1 / releaseSteps;
+    }
+    // For sine waves, the envelope eventually reaches an average power of
```
hoch: Add one line break at l.171. Also not sure if this comment is at the right place. It explains what's happening at l.182. No?

audio-worklet/src/noisegate-audio-worklet.js
```diff
@@ -0,0 +1,216 @@
...
+    // a^2 / 2. Sine waves are therefore scaled back to the original
+    // amplitude, but other waveforms or constant sources can only be
+    // approximated.
+    let scaledEnvelopeValue;
+    let weight;
+
+    // Compute an array of weights between 0 and 1 which will be multiplied with
```
hoch: The comment should explain "why" not about "what". This comments describe what is happening below.

PR #115, opened 2017-08-15
Added three master effects to the poly-synth

samhincks:
In this PR, three master effects have been added to the poly-synth: convolution, feedback delay, and bitcrushing. The PR modifies 4 files:
- poly-synth/index.html: added one dependency, and changed the GUI
- poly-synth/src/poly-synth-demo.js: added GUI components for master effects and impulse response loading.
- poly-synth/src/poly-synth.js: added the audio graph supporting effects and controls for changing effects settings.
- poly-synth/src/poly-synth-voice.js: removed an extraneous blank line!

The demo can be found [here](http://rawgit.com/GoogleChrome/web-audio-samples/8a4615e3124b83c4f783f4c188a8d302ab1f943b/poly-synth/index.html). As I evaluate the demo online, I notice that it takes a couple seconds to load. This is because it takes time to load the impulse response buffers. I may want to consider the number of impulse response buffers so that it loads quicker.

### DIFF
--- a/poly-synth/src/poly-synth-demo.js
+++ b/poly-synth/src/poly-synth-demo.js
@@ -22,55 +22,55 @@ class PolySynthDemo {
   /**
    * @constructor
    * @param {AudioContext} context The audio context.
-   * @param {String} gainADSRId The ID of the container for gain ADSR elements
-   *                            (e.g. <div>).
-   * @param {String} filterADSRId The ID of the container for filter
-   *                              ADSR elements.
-   * @param {String} bitcrusherId The ID of the container for bitcrusher
-   *                              parameter elements.
-   * @param {String} convolverId The ID of the container for convolver
-   *                             elements.
-   * @param {String} convolverSelectorId The ID of the <select> element for
-   *                             convolver impulse response options.
-   * @param {String} feedbackId The ID of the container for feedback delay
-   *                             elements.
+   * @param {Object} identifiers
+   * @param {String} identifiers.gainEnvelopeDivId The ID of the container for
+   *                                               gain envelope elements.
+   * @param {String} identifiers.filterDivId The ID of the container for
+   *                                         filter elements.
+   * @param {String} identifiers.filterEnvelopeDivId The ID of the container for
+   *                                                 filter envelope elements.
+   * @param {String} identifiers.bitcrusherDivId The ID of the container for
+   *                                             bitcrusher parameter elements.
+   * @param {String} identifiers.reverbDivId The ID of the container for
+   *                                         reverb elements.
+   * @param {String} identifiers.reverbSelectorId The ID of the <select>
+   *                                              element for impulse
+   *                                              response options.
+   * @param {String} identifiers.delayDivId The ID of the container for
+   *                                        feedback delay elements.
+   * @param {String} identifiers.volumeDivId The ID of the container for
+   *                                         the volume control.
    */
-  constructor(
-      context, gainADSRId, filterADSRId, bitcrusherId, convolverId,
-      convolverSelectorId, feedbackId) {
+  constructor(context, identifiers) {
     this.context_ = context;
-    this.gainADSRId_ = gainADSRId;
-    this.filterADSRId_ = filterADSRId;
-    this.bitcrusherId_ = bitcrusherId;
-    this.convolverId_ = convolverId;
-    this.convolverSelectorId_ = convolverSelectorId;
-    this.feedbackId_ = feedbackId;
-
+    this.reverbSelectorId_ = identifiers.reverbSelectorId;
     this.impulseResponseUrls_ = this.getImpulseResponseUrls_();
 
     this.masterGain_ = new GainNode(this.context_, {gain: 0.5});
+    this.polySynth_ =
+        new PolySynth(this.context_);
+    this.polySynth_.output.connect(this.masterGain_)
+        .connect(this.context_.destination);
 
+    // The QWERTY keyboard is defined in qwerty-hancock.min.js.
+    this.keyboard_ = new QwertyHancock(
+        {id: 'keyboard', width: 600, height: 150, octaves: 2});
+    this.keyboard_.keyDown = this.keyDown.bind(this);
+    this.keyboard_.keyUp = this.keyUp.bind(this);
+    
+    this.initializeAmplifierEnvelopeGUI_(identifiers.gainEnvelopeDivId);
+    this.initializeFilterGUI_(identifiers.filterDivId);
+    this.initializeFilterEnvelopeGUI_(identifiers.filterEnvelopeDivId);
+    this.initializeBitcrusherGUI_(identifiers.bitcrusherDivId);
+    this.initializeDelayGUI_(identifiers.delayDivId);
+    this.initializeReverbGUI_(identifiers.reverbDivId, this.reverbSelectorId_);
+    this.initializeMasterVolumeGUI_(identifiers.volumeDivId);
+    
     this.loadImpulseResponses(this.impulseResponseUrls_)
         .then((impulseResponseBuffers) => {
           this.impulseResponseBuffers_ = impulseResponseBuffers;
-          
-          this.polySynth_ =
-              new PolySynth(this.context_, impulseResponseBuffers[0]);
-          this.polySynth_.output.connect(this.masterGain_)
-              .connect(this.context_.destination);
-
-          // The QWERTY keyboard is defined in qwerty-hancock.min.js.
-          this.keyboard_ = new QwertyHancock(
-              {id: 'keyboard', width: 600, height: 150, octaves: 2});
-          this.keyboard_.keyDown = this.keyDown.bind(this);
-          this.keyboard_.keyUp = this.keyUp.bind(this);
-          
-          this.initializeAmplifierEnvelopeGUI_(this.gainADSRId_);
-          this.initializeFilterEnvelopeGUI_(this.filterADSRId_);
-          this.initializeBitcrusherGUI_(this.bitcrusherId_);
-          this.initializeConvolverGUI_(
-              this.convolverId_, this.convolverSelectorId_);
-          this.initializeFeedbackDelayGUI_(this.feedbackId_);
+          this.polySynth_.setConvolverBuffer(this.impulseResponseBuffers_[0]);
+          document.getElementById(this.reverbSelectorId_).disabled = false;
         });
   }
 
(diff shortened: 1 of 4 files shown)


poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
-    this.keyboard_ = new QwertyHancock(
-        {id: 'keyboard', width: 600, height: 150, octaves: 2});
-
-    this.keyboard_.keyDown = this.keyDown.bind(this);
-    this.keyboard_.keyUp = this.keyUp.bind(this);
-  }
-
-  /**
-   * Initialize GUI components.
    * @param {String} gainADSRId The ID of the container for gain ADSR elements
```
hoch: s/ADSR/Envelope/

I vaguely remember that I pointed this out before.

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
    *                            (e.g. <div>).
    * @param {String} filterADSRId The ID of the container for filter
    *                              ADSR elements.
+   * @param {String} bitcrusherId The ID of the container for bitcrusher
```
hoch: Also it is a bit confusing that the term `fooId` is actually referring to a DOM element. `bitchrusher` isn't a DOM element.

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
+   *                             elements.
+   * @param {String} convolverSelectorId The ID of the <select> element for
+   *                             convolver impulse response options.
+   * @param {String} feedbackId The ID of the container for feedback delay
+   *                             elements.
    */
-  initializeGUI(gainADSRId, filterADSRId) {
+  constructor(
+      context, gainADSRId, filterADSRId, bitcrusherId, convolverId,
+      convolverSelectorId, feedbackId) {
```
hoch: A constructor with 7 arguments is not a nice thing to look at. :)

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
+  }
+
+  async loadSound(url) {
+    const response = await fetch(url);
+    const sound = await response.arrayBuffer();
+    return this.context_.decodeAudioData(sound);
+  }
+
+  initializeAmplifierEnvelopeGUI_(containerId){
     let masterGainSlider_ =
```
hoch: Nit: the master gain is not part of the amp envelope.

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
+        bitcrusherReductionSlider_.enable();
+      }
+    }
+  }
+
+  initializeConvolverGUI_(containerId, selectorId) {
+    let convolverWetnessSlider_ = new ParamController(
+        containerId, this.polySynth_.setConvolverWetness.bind(this.polySynth_),
+        {
+          name: 'Wetness',
```
hoch: How about.. "mix"?

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
+          type: 'range',
+          min: 0,
+          max: 1,
+          step: 0.1,
+          default: this.polySynth_.feedbackDelayTime
+        });
+
+    let feedbackGainSlider_  = new ParamController(
+        containerId, this.polySynth_.setFeedbackGain.bind(this.polySynth_), {
+          name: 'Gain',
```
hoch: Gain -> Feedback

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
+          type: 'range',
+          min: 0,
+          max: 1,
+          step: 0.1,
+          default: this.polySynth_.feedbackWetness
+        });
+
+    let feedbackDelaySlider_ = new ParamController(
+        containerId, this.polySynth_.setFeedbackDelayTime.bind(this.polySynth_), {
+          name: 'Delay',
```
hoch: Delay -> DelayTime

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
+          type: 'range',
+          min: 0,
+          max: 1,
+          step: 0.1,
+          default: this.polySynth_.feedbackGain
+        });
+  }
+
+  getImpulseResponseUrls_() {
+    let impulseResponseUrls =
```
hoch: Basically every time you call this method, an array with a bunch of strings will be created. Is that what we want?

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
           max: this.polySynth_.maxCutoff,
           step: 1,
-          default: this.polySynth_.getParameters().filterCutoff,
+          default: this.polySynth_.getParameters().filterCutoff
         });
 
     let filterQSlider_ = new ParamController(
-        filterADSRId, this.polySynth_.setParameter.bind(this.polySynth_), {
+        containerId, this.polySynth_.setParameter.bind(this.polySynth_), {
           name: 'Q (dB)',
```
hoch: Q is not in dB.

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
           max: this.polySynth_.maxRelease,
           step: 0.01,
-          default: this.polySynth_.getParameters().filterRelease,
+          default: this.polySynth_.getParameters().filterRelease
         });
 
     let filterDetuneSlider_ = new ParamController(
-        filterADSRId, this.polySynth_.setParameter.bind(this.polySynth_), {
+        containerId, this.polySynth_.setParameter.bind(this.polySynth_), {
           name: 'Detune Amount',
```
hoch: Detune Amount -> Filter Amount

poly-synth/src/poly-synth-demo.js
samhincks: In the spec it says: "Controls how peaked the response will be at the cutoff frequency. A large value makes the response more peaked. Please note that for this filter type, this value is not a traditional Q, but is a resonance value in decibels."

poly-synth/src/poly-synth-demo.js
samhincks: Yes, I think separating this into its own method is cleaning than leaving this massive array initializer in the constructor.

poly-synth/src/poly-synth-demo.js
samhincks: Under the hood it isn't, but I placed it under the amplifier section of the GUI. Do you think that should be changed? I like having the master volume in this part of the GUI but unlike the other sliders in this section the master volume can be switched for realtime effect even if the voice has already been constructed.

poly-synth/src/poly-synth-demo.js
samhincks: Perhaps I should bundle the GUI IDS together into an object

poly-synth/src/poly-synth-demo.js
samhincks: You pointed it out in the HTML file. I should have changed its name in all the JS files too.

poly-synth/src/poly-synth-demo.js
samhincks: With this name, how do you know that 1 refers to more of the wet signal?

poly-synth/src/poly-synth.js
```diff
@@ -60,6 +62,22 @@ class PolySynth {
     this.minQ = -50;
     this.maxQ = 50;
 
+    // The bitcrusher defaults non-information reducing parameters.
```
rtoy: Don't remember right now, but why aren't the bitcrusher nodes defaults set this way already?

poly-synth/src/poly-synth.js
```diff
@@ -72,13 +90,101 @@ class PolySynth {
...
+    bitcrusherOutputNode.connect(convolverInputNode);
+    convolverOutputNode.connect(this.output);
+  }
+
+  createFeedbackGraph_(inputNode) {
+    // The sum of the gain of the wet and dry signal is always 1, and the output
+    // always routes through both nodes.
+    this.feedbackWetSignal_ =
+        new GainNode(this.context_, {gain: this.feedbackWetness});
+    this.feedbackDrySignal_ =
```
rtoy: I find these names confusing when thinking about the graph.  Maybe feedbackWetSignalGain and feedbackDrySignalGain?  That tells me these are the gains for the different signals, not the signals themselves.Maybe.

poly-synth/src/poly-synth.js
```diff
@@ -72,13 +90,101 @@ class PolySynth {
...
   }
-  
+
+  createBitcrusherGraph_(inputNode) {
+    this.bitcrusher = new Bitcrusher(
+        this.context_,
+        {bitDepth: this.maxBitDepth, reduction: this.minReduction});
+
+    // Only one of |this.activeBitcrusherRoute_| and
+    // |this.bypassBitcrusherRoute_| has a non-zero gain
```
rtoy: This isn't really true, right? You do linear-ramp these so at some point both are non-zero.

rtoy:
Got these errors in the console when starting the demo. Probably want to fix that:

> Uncaught TypeError: Cannot read property 'release' of undefined
    at PolySynth.releaseVoice (poly-synth.js:306)
    at PolySynthDemo.keyUp (poly-synth-demo.js:415)
    at D (qwerty-hancock.min.js:23)
    at qwerty-hancock.min.js:23
poly-synth.js:306 Uncaught TypeError: Cannot read property 'release' of undefined
    at PolySynth.releaseVoice (poly-synth.js:306)
    at PolySynthDemo.keyUp (poly-synth-demo.js:415)
    at D (qwerty-hancock.min.js:23)
    at qwerty-hancock.min.js:23

rtoy:
Hmm.  Can't reproduce those errors any more.

poly-synth/src/poly-synth-demo.js
hoch: It is `mix` parameter of `Reverb` component. So the bigger mix value means you get more reverberation. I think it's intuitive - also it is the convention.

hoch:
Oh, that's not good. Does qwerty-hancock do any async stuff?

poly-synth/src/poly-synth-demo.js
hoch: Oh. That's nice. FYI, many synth use arbitrary ranges for Q.

poly-synth/src/poly-synth-demo.js
hoch: Not sure why it's difficult to move this out of this method, but it is just a nit. This is not a reusable class, so I am fine with it.

poly-synth/src/poly-synth-demo.js
```diff
@@ -22,163 +22,368 @@ class PolySynthDemo {
...
+
+    // Sound is not processed by the bitcrusher if in bypass mode.
+    document.getElementById('bitcrusherBypassButton').onclick = (event) => {
+      // Only one of |this.polySynth_.activeBitcrusherRoute_| and
+      // |this.polySynth_.bypassBitcrusherRoute_| has a non-zero gain.
+      if (event.target.textContent === 'Active') {
+        event.target.textContent = 'Bypassed';
+
+        // The change is scheduled slightly into the future to avoid glitching.
+        const t = this.context_.currentTime + 0.01;
```
samhincks: You certainly hear that a swapping is occurring but its not a pronounced glitch as it is if the swap occurs immediately

poly-synth/src/poly-synth.js
samhincks: They are set that way, but maximum and minimum values still need to be set explicitly for the sliders. I took out the comment since it is unnecessary.

samhincks:
In the latest [commit](https://github.com/GoogleChrome/web-audio-samples/pull/115/commits/f3c8ad0a28c6803218c3088a3f8482e7592e1c76),
- the convolver effect has been renamed reverb, and associated variable names have changed
- the feedback effect has been renamed delay, and associated variable names have changed 
- html controls now have a fixed width
- the synthesizer is initialized before the convolver buffers load
- Filter and Master volume have separate containers
- other modifications based on review and in person discussion.

The demo can be found [here](http://rawgit.com/GoogleChrome/web-audio-samples/f3c8ad0a28c6803218c3088a3f8482e7592e1c76/poly-synth/index.html).

I have not identified the source of the bug described above. But I'll keep investigating, and file an issue for it if it seems to persist. Perhaps if it occurs the first time a page is loaded but not later, the cache prevents it from happening? The [Qwerty keyboard ](https://raw.githubusercontent.com/stuartmemo/qwerty-hancock/master/dist/qwerty-hancock.js) creates listeners for listening to mouse and keyboard input but it doesn't do any other asynchronous operations from what I can tell.

PR #134, opened 2017-12-15
Fix stress-test bug

Jedipedia:
Currently, the stress-test demo only works in Chrome, not in Firefox. Firefox throws a ```ReferenceError: event is not defined``` because the function argument is called ```clickEvent```, not ```event```. I'm not sure why the code works in Chrome, but with this change, it will work in both browsers.


Jedipedia:
I signed it!

hoch:
Thanks for the fix!

PR #149, opened 2018-05-14
AW Design Pattern 2: RingBuffer in AWP

hoch:
PTAL.

The associated section: https://docs.google.com/document/d/1R78PuADY671xGMcSwcw7SX7DVFg1K0h4jOFKqWUxdsc/edit#bookmark=id.vj4uufhcm1r9


audio-worklet/design-pattern/awn-ring-buffer/RingBufferWorkletProcessorKernel.cc
```diff
@@ -0,0 +1,46 @@
...
+
+// A multi-channel processing kernel. To handle multiple inputs or outputs,
+// simply use multiple instances of this kernel. The design assumes:
+//   1. The kernel size (processing frame size) is static after construction.
+//   2. The channel count can be changed dynamically.
+//   3. The given channel count must be smaller than the size of the array
+//      being passed in.
+class RingBufferWorkletProcessorKernel {
+ public:
+  RingBufferWorkletProcessorKernel(unsigned kernel_buffer_size)
```
rtoy: Naming is a bit confusing.  `kernel_buffer_size` is basically the size of the FIFO?

audio-worklet/design-pattern/awn-ring-buffer/ring-buffer-worklet-processor.js
```diff
@@ -0,0 +1,84 @@
...
+   * @return {Boolean} Active source flag.
+   */
+  process(inputs, outputs, parameters) {
+    // Use the 1st input and output only to make the example simpler. |input|
+    // and |output| here have the similar structure with the AudioBuffer
+    // interface. (i.e. An array of Float32Array)
+    let input = inputs[0];
+    let output = outputs[0];
+
+    // AudioWorkletProcessor always gets 128 frams in and 128 frames out. Here
```
rtoy: "128 frams"

hoch: Done.

audio-worklet/design-pattern/awn-ring-buffer/RingBufferWorkletProcessorKernel.cc
hoch: Although the class name contains "RingBuffer" in it, this class does not implement the ring buffer. Yes, technically the `kernel_buffer_size` is equal to the size of RingBuffer - that's because the user function (the kernel) requires a certain buffer size, and the RingBuffer is there to accommodate such buffer size within AudioWorkletProcessor.

Perhaps taking "RingBuffer" out of the class name would be better?

rtoy: Then it becomes WorkletProcessorKernel, which is kind of generic.  Maybe WorkletProcessorKernelWithRingBuffer (or WithFIFO)?

I'll leave the final choice to you.

hoch:
Here's summary of the final change:
1. C++ class is renamed to: `AWPKernelWithVariableBufferSize`.
2. The generated WASM glue code was renamed: `audioworklet.wasmmodule.js`. I think this is okay to have a generic name because it is a big mixture of WASM code + helpers.

PR #151, opened 2018-05-30
Fix discrepancy between module name in index.html & Makefile

habemus-papadum:
Small typo.

-----------
Cool stuff -- thanks for providing these samples.

### DIFF
--- a/audio-worklet/design-pattern/awn/Makefile
+++ b/audio-worklet/design-pattern/awn/Makefile
@@ -9,4 +9,4 @@ build: $(DEPS)
 		--post-js audio-worklet-processor.js
 
 clean:
-	@rm -f audio-worklet-module.js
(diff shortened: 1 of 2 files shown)


hoch:
Well, I have not announced it yet so I did not expect someone will take a look at it. :)
Thanks for fixing this! Will merge once you singed the CLA.

habemus-papadum:
I signed the CLA, but I realized that the change would be better to make in the Makefile (in my PR the clean target is wrong)  I'm making the edits from the browser via Github, so I am going to create another PR.  I kill this one in a day or so.

habemus-papadum:
I just added a second commit to this PR -- sorry for being sloppy not merging the commits, I was curious how far I could go  using only the GH web UI.  I believe you can squash the commits easily via the GH web PR ui, if that is acceptable.  cheers

PR #155, opened 2018-07-10
Finalizing AudioWorklet Design Pattern

hoch:
PTAL.

- Rename files to clarify its use case.
- Use ES6 import/export for the better dependency handling.
- Expose links to design pattern examples in the index page.


audio-worklet/design-pattern/lib/em-es6-module.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
+ */
+
+// EXPORT_ES6 option does not work as described at
+// https://github.com/kripken/emscripten/issues/6284, so we have to
+// manually add this by '--post-js' setting when the emcc compilation.
```
rtoy: "manually add...emcc compilation" is phrased in a funny way that I don't know what it means.

audio-worklet/design-pattern/lib/em-es6-module.js
```diff
@@ -0,0 +1,10 @@
...
+export default Module;
```
rtoy: Funny EOL character here?

audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -256,3 +250,11 @@ class RingBuffer { // eslint-disable-line no-unused-vars
...
   }
 } // class RingBuffer
+
+
+export {
+  MAX_CHANNEL_COUNT,
+  RENDER_QUANTUM_FRAMES,
+  HeapAudioBuffer,
+  RingBuffer,
+};
```
rtoy: Funny EOL character?

audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -1,27 +1,20 @@
 /**
- * @fileOverview WA2 library. A set of utilities to support WASM operations
- * for the WebAudio. (WebAudio and WebAssembly => WAWA => WA2)
+ * Copyright (c) 2018 The Chromium Authors. All rights reserved.
+ * Use of this source code is governed by a BSD-style license that can be
+ * found in the LICENSE file.
```
rtoy: While valid, should we not use '//'-style comments instead of C-style?  Here and elsewhere.

audio-worklet/design-pattern/shared-buffer/shared-buffer-worker.js
```diff
@@ -119,6 +126,15 @@ function initialize(options) {
     CONFIG.channelCount = options.channelCount;
   }
 
+  if (!self.SharedArrayBuffer) {
+    postMessage({
+      message: 'WORKER_ERROR',
+      detail: `SharedArrayBuffer is not supported. See the article above to
```
rtoy: This message doesn't really help, does it?  Where is the "article above"?  Maybe link (with shortened URL) to the actual article?

audio-worklet/design-pattern/lib/em-es6-module.js
hoch: Done.

audio-worklet/design-pattern/lib/em-es6-module.js
hoch: Done. VSCode doesn't do this for me by default. :(

audio-worklet/design-pattern/lib/wasm-audio-helper.js
hoch: I think we should follow what DevRel team does. See the example here: https://github.com/GoogleChromeLabs/audioworklet-polyfill/blob/master/src/realm.js

FYI: Based on the suggestion from DevRel, I am changing the license to Apache 2.

audio-worklet/design-pattern/shared-buffer/shared-buffer-worker.js
hoch: My mistake - now the page link is ready so I can add it in here. I also prefer the actual link than the shortened one.

audio-worklet/design-pattern/lib/em-es6-module.js
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+// EXPORT_ES6 option does not work as described at
+// https://github.com/kripken/emscripten/issues/6284, so we have to
+// manually add this by '--post-js' setting when the Emscripten compilation.
+export default Module;
```
frantic0: This seems to work well when you don't compile with EMSCRIPTEN with the `-s MODULARIZE=1'

Not so, otherwise

hoch: That's why the Makefile does not use that command line option:
https://github.com/GoogleChromeLabs/web-audio-samples/blob/master/audio-worklet/design-pattern/wasm/Makefile

Are you suggesting the current code does not compile?

frantic0: Sorry, my comment was unclear. The code compiles well as is. 

If I understood correctly, that set of command line options is the only one that guarantees the correct loading of the wasm module to the AudioWorkletGlobalScope?

I have been trying a compilation with ``-s MODULARIZE_INSTANCE=1`` and ``-s EXPORT_NAME="'<module_name>'"`` but I'm getting a DOMException when loading the worklet.

hoch: AudioWorkletGlobalScope might be missing some APIs that are available in the main global or WorkerGlobalScope.

`--post-js` above was sort of a workaround because `EXPORT_ES6`option did not work for me. See https://github.com/emscripten-core/emscripten/issues/6284.

PR #158, opened 2018-08-16
Fix #156: Create new convolver when effect changes

### LINKED ISSUE #156: ConvolverNode buffer can only be set once
According to https://webaudio.github.io/web-audio-api/#dom-convolvernode-buffer, the `buffer` attribute can  only be set to a non-null value once.  Chrome is now enforcing this and produces the required error.  The [shiny-drum-machine demo](https://googlechromelabs.github.io/web-audio-samples/samples/audio/shiny-drum-machine.html) needs to be fixed.  There may be others.

rtoy:
It is illegal to set the convovler buffer to a new value once it's
already been set.  Thus, we need to create a new convolver whenever
the effect changes.  Disconnect the old convolver from downstream
nodes and connect up the new convolver to the old downstream node.

### DIFF
--- a/samples/audio/shiny-drum-machine.html
+++ b/samples/audio/shiny-drum-machine.html
@@ -470,11 +470,6 @@ function init() {
     effectLevelNode.gain.value = 1.0; // effect level slider controls this
     effectLevelNode.connect(masterGainNode);
 
-    // Create convolver for effect
-    convolver = context.createConvolver();
-    convolver.connect(effectLevelNode);
-
-
     var elKitCombo = document.getElementById('kitcombo');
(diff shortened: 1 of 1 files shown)


rtoy:
Also need to update convolution-effects.html, simple.html.

samples/audio/shiny-drum-machine.html
```diff
@@ -945,8 +940,16 @@
     theBeat.effectIndex = index;
     effectDryMix = impulseResponseInfoList[index].dryMix;
     effectWetMix = impulseResponseInfoList[index].wetMix;            
+    
+    if (convolver != undefined) {
```
hoch: `if (convolver) {`

Isn't this good enough?

PR #185, opened 2019-07-11
Add classes for modeling the audio graph

gaopinghuang0:
The third commit adds more method comments to the first commit.


geiger/src/graph/Graph.js
```diff
@@ -51,6 +54,10 @@ export default class Graph extends Events {
     this._in = {};
   }
 
+  /**
+   * Add cell to the graph based on its type.
+   * @param {!Cell} cell
```
hoch: What's `Cell`? This seems to be something other than `Node`.

geiger/src/graph/Graph.js
```diff
@@ -62,6 +69,10 @@ export default class Graph extends Events {
     }
   }
 
+  /**
+   * Add a node to the graph.
+   * @param {!Node} node
```
hoch: Where is `Node` defined? (as in `@typedef`)

geiger/src/graph/Graph.js
```diff
@@ -129,6 +144,10 @@ export default class Graph extends Events {
     this._shouldRedraw();
   }
 
+  /**
+   * Remove an edge to the graph by its id.
+   * @param {!string} edgeId
```
hoch: Let's define a data type for these UUIDs. I am using `GraphNodeId` in my JS code which extends `string`. I suggest you to do the same.

hoch: Hmm. This is an edge not a node. I'll have to think about it a bit.

geiger/src/graph/NodeWithPort.js
```diff
@@ -242,6 +242,11 @@ export default class NodeWithPort extends Node {
     return ports;
   }
 
+  /**
+   * Validate ports created and then store.
+   * @param {!Array<Port>} ports
```
hoch: Where is `Port` is defined?

geiger/src/graph/NodeWithPort.js
```diff
@@ -242,6 +242,11 @@ export default class NodeWithPort extends Node {
...
+   * @return {!Collection}
```
hoch: Also what is `Collection`? Are these are included/imported from somewhere else?

geiger/src/graph/NodeWithPort.js
```diff
@@ -242,6 +242,11 @@ export default class NodeWithPort extends Node {
...
+   */
   _createPortData(ports) {
     const err = this._validatePorts(ports);
```
hoch: Always spell things out if possible. (err => error)

geiger/src/graph/Graph.js
gaopinghuang0: `Cell` is a base class for `Node` and `Edge`.

geiger/src/graph/Graph.js
gaopinghuang0: Since `Node` is a module already imported via `import ...`, I simply ignore the `@typedef`.

geiger/src/graph/NodeWithPort.js
gaopinghuang0: `Port` is defined at the beginning of the file, right above the class comments.

geiger/src/graph/NodeWithPort.js
gaopinghuang0: Since `Collection` is a module already imported via `import ...`, I simply ignore the `@typedef`.

geiger/src/graph/NodeWithPort.js
gaopinghuang0: Sure.

geiger/src/graph/Cell.js
```diff
@@ -0,0 +1,37 @@
...
+/** @typedef {import('../graph/Graph').default} Graph */
+
+/**
+ * A Cell could be Node or Edge. They all have property `graph`.
+ * @class Cell
+ */
+export default class Cell {
+  constructor() {
+    this.graph = null;
+  }
```
hoch: I think this abstraction is a bit weird. Can a `cell` exist without a graph? If not, wouldn't it be better to have `graph` as a constructor argument?

geiger/src/graph/Cell.js
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * A shortcut to add cell to graph.
+   * @param {!Graph} graph
+   * @return {!Cell}
+   */
+  addTo(graph) {
```
hoch: ...then we don't need this method at all.

geiger/src/graph/Edge.js
```diff
@@ -0,0 +1,141 @@
...
+  constructor(message, type) {
+    super();
+
+    this.id = null;
+    this.type = type;
+
+    /** @type {!Source} */
+    this.source = null;
+    /** @type {!Target} */
+    this.target = null;
```
hoch: Let's use the term `destination` than `target` to make it consistent.

geiger/src/graph/Edge.js
```diff
@@ -0,0 +1,141 @@
...
+    this._points = points;
+  }
+
+  toString() {
+    return [this.id, JSON.stringify(this.source),
+      JSON.stringify(this.target)].join(',');
+  }
+
+  shouldRender() {
+    return !!this._points;
```
hoch: Let's be clear about this. Perhaps `this._points.length > 0`?

geiger/src/graph/Edge.js
```diff
@@ -0,0 +1,141 @@
...
+   * @param {!ConnectionMessage} message
+   * @param {!EdgeTypes} type
+   */
+  constructor(message, type) {
+    super();
+
+    this.id = null;
+    this.type = type;
+
+    /** @type {!Source} */
```
hoch: Shouldn't this be `Node` instead of `Source`?

geiger/src/graph/Edge.js
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * @typedef {NodesConnection | NodeParamConnection} ConnectionMessage
+ */
+
+/**
+ * @typedef {Object} Source
+ * @property {string} id - The node id.
+ * @property {string} port - The port id.
+ * @typedef {Source} Target
+ */
```
hoch: I don't think this abstraction is useful.

geiger/src/graph/Edge.js
```diff
@@ -0,0 +1,141 @@
...
+   */
+  constructor(message, type) {
+    super();
+
+    this.id = null;
+    this.type = type;
+
+    /** @type {!Source} */
+    this.source = null;
+    /** @type {!Target} */
```
hoch: Ditto.

geiger/src/graph/Edge.js
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * @typedef {Object} NodeParamConnection
+ * @property {string} sourceNodeId
+ * @property {string} destinationNodeId
+ * @property {string} fromChannel
+ * @property {string} destinationParamName
+ */
+
+/**
+ * @typedef {NodesConnection | NodeParamConnection} ConnectionMessage
```
hoch: You might want to revisit this based on the event/data structure in the chromium CL.

geiger/src/graph/Edge.js
```diff
@@ -0,0 +1,141 @@
...
+ */
+
+/**
+ * @typedef {Object} Source
+ * @property {string} id - The node id.
+ * @property {string} port - The port id.
```
hoch: Does `port` also have an ID?

geiger/src/graph/Graph.js
```diff
@@ -0,0 +1,226 @@
...
+    // Cache the removed nodes/edges in current batch.
+    this._removed_nodes = [];
+    this._removed_edges = [];
+
+    /**
+     * For each node ID, keep track of all out-bound edge IDs.
+     * @type {!Object<string, !Object<string, true>>}
+     * @private
+     */
+    this._out = {};
```
hoch: `_outbound_edge_map`

geiger/src/graph/Graph.js
```diff
@@ -0,0 +1,226 @@
...
+
+    /**
+     * For each node ID, keep track of all in-bound edge IDs.
+     * @type {!Object<string, !Object<string, true>>}
+     * @private
+     */
+    this._in = {};
```
hoch: Ditto.

geiger/src/graph/Graph.js
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  /**
+   * Add cell to the graph based on its type.
+   * @param {!Cell} cell
+   */
+  addCell(cell) {
+    if (!cell.graph) {
+      cell.graph = this;
```
hoch: This is not a good practice: Cell can add a graph into it, and also graph add a cell into itself. Let's make this a one-directional operation.

When you do `new Cell(graph)`, all of things should be triggered.

geiger/src/graph/label.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+/**
+ * Computes the id for the input of node.
+ * @param {string | number} contextId
+ * @param {string} nodeId
+ * @param {string | number} portIndex
+ * @return {string}
+ */
+export const computeInPortId = (contextId, nodeId, portIndex) => {
```
hoch: We're not computing. It generates a string. Let's rename the functions here.

geiger/src/util/Collection.js
```diff
@@ -0,0 +1,144 @@
+import {isArray} from 'util';
+import defaults from 'lodash.defaults';
```
hoch: Uh-oh. Do we have a dependency on `lodash`?

geiger/src/util/Collection.js
```diff
@@ -0,0 +1,144 @@
...
+    return this;
+  }
+
+  /**
+   * Add an array of items, each item should have id.
+   * @param {Array<T>} items
+   */
+  addItems(items) {
+    items.forEach((item) => {
+      if (isObject(item)) {
```
hoch: Perhaps an assert on `id` is required?

geiger/src/graph/Cell.js
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+/** @typedef {import('../graph/Graph').default} Graph */
+
+/**
+ * A Cell could be Node or Edge. They all have property `graph`.
+ * @class Cell
+ */
+export default class Cell {
```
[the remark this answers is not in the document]
gaopinghuang0: I removed the class Cell.

geiger/src/graph/Cell.js
gaopinghuang0: Removed.

geiger/src/graph/Cell.js
gaopinghuang0: Removed.

geiger/src/graph/Edge.js
gaopinghuang0: An `Edge` is pointing to a specific port of a `Node`. Each port has an ID. Perhaps we can use `this.source` to point to the node, then add another property to store the port ID.

Also, when creating an edge, is it good to pass a `graph` into its constructor? So that we can find the node based on the node ID from the `graph`.

geiger/src/graph/Edge.js
gaopinghuang0: Yes.

geiger/src/util/Collection.js
gaopinghuang0: Just one function called `defaults`. I can rewrite it.

geiger/src/util/Collection.js
gaopinghuang0: If an item has no `id`,  an error will be thrown inside `addItem()`. That's why I didn't add an assert here. But I can add here.

geiger/src/graph/Graph.js
gaopinghuang0: Removed.

geiger/src/graph/label.js
gaopinghuang0: Renamed.

geiger/src/graph/Graph.js
gaopinghuang0: Done.

geiger/src/graph/Node.js
```diff
@@ -0,0 +1,95 @@
...
+   * @param {!Point} position
+   * @return {!Cell} - To be chainable
+   */
+  setPosition(position) {
+    this._position = position;
+    return this;
+  }
+
+  shouldRender() {
+    return !!this._position;
```
hoch: Let's not use `!!` unless if it's absolutely necessary.

What's your intention here? Perhaps it's better to keep a separate flag like `_needsRender`?

geiger/src/graph/NodeWithPort.js
```diff
@@ -0,0 +1,282 @@
...
+  getPortById(id) {
+    return this._ports.get(id);
+  }
+
+  /**
+   * @param {!PortTypes} type
+   * @return {Array<Port>}
+   */
+  getPortsByType(type) {
+    return this._ports.filter((port) => port.type === type);
```
hoch: nit: `(port)` can be just `port`.

geiger/src/graph/NodeWithPort.js
```diff
@@ -0,0 +1,282 @@
...
+      if (!this._isValidPortId(p.id)) {
+        errorMessages.push('port id is required');
+      }
+    });
+
+    return errorMessages;
+  }
+
+  _isValidPortId(id) {
+    return id !== null && id !== undefined && !isObject(id);
```
hoch: We might not need this kinda check, because the Browser Protocol layer will do the validation for us.

geiger/src/graph/label.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+/**
+ * Computes the id for a node.
+ * @param {string | number} contextId
+ * @param {string} nodeId
+ * @return {string}
+ */
+export const computeNodeId = (contextId, nodeId) => {
+  return 'c' + contextId + 'n' + nodeId;
```
hoch: Ditto: use the template literal instead of concat. See below.

geiger/src/graph/label.js
```diff
@@ -0,0 +1,77 @@
...
+};
+
+/**
+ * Computes the id of the edge using port ids of source and destination.
+ * @param {string} sourcePortId
+ * @param {string} destinationPortId
+ * @return {string}
+ */
+export const computeEdgeId = (sourcePortId, destinationPortId) => {
+  return sourcePortId + '|' + destinationPortId;
```
hoch: What's `|`? Also let's use the template literal.

geiger/src/graph/NodeWithPort.js
gaopinghuang0: Linter complains if I remove the parentheses.

gaopinghuang0:
In the last commit, I updated the API of Graph to match the events defined in WebAudioModel.

Some notes:
1. Since the AudioParam is added after the AudioNode is created, the `NodeWithPort` class has a big change about how to compute node size and set port position. (Previously, it was assumed that `NodeCreated` message contains all the AudioParams related to the Node. So the computation is quite different.)
2. The types of messages and some common objects are put in a file called `jsdoc.types.js`, which can be shared by multiple files via `// /<reference path="../jsdoc.types.js" />`. Now this line is put at the beginning of most files, right after the import statements.

geiger/src/graph/Edge.js
```diff
@@ -0,0 +1,104 @@
...
+
+  let destinationPortId;
+  if (type === EdgeTypes.NODE_TO_NODE) {
+    destinationPortId = generateInputPortId(message.destinationId,
+        message.destinationInputIndex);
+  } else if (type === EdgeTypes.NODE_TO_PARAM) {
+    destinationPortId = generateParamPortId(message.destinationId,
+        message.destinationParamId);
+  } else {
+    throw new Error('Unknown edge type: ' + type);
```
hoch: I think we want to print out console messages and ignore invalid behavior. This exception will stop the devtool panel from working.

geiger/src/graph/Graph.js
```diff
@@ -0,0 +1,283 @@
...
+  /**
+   * @constructor
+   * @param {!string} contextId
+   */
+  constructor(contextId) {
+    super();
+
+    this.contextId = contextId;
+
+    this._nodes = new Collection(null, {trackExit: true});
```
hoch: Let's annotate the types of `_nodes` and `_edges`.

geiger/src/graph/Graph.js
```diff
@@ -0,0 +1,283 @@
...
+   */
+  removeNode(nodeId) {
+    this._nodes.remove(nodeId);
+    this._removeEdgesOfNode(nodeId);
+    this._shouldRedraw();
+  }
+
+  /**
+   * Add a param to the node.
+   * @param {!ParamCreationMessage} message
```
hoch: `Message` feels like a bunch of string needs to be printed. How about `ParamCreationData`?

geiger/src/graph/Graph.js
```diff
@@ -0,0 +1,283 @@
...
+   */
+  removeParam(paramId, nodeId) {
+    const node = this.getNodeById(nodeId);
+    node.removeParamPort(paramId);
+    this._shouldRedraw();
+  }
+
+  /**
+   * Add a Node-to-Node connection to the graph.
+   * @param {!NodesConnectionMessage} message
```
hoch: Ditto. Let's use `Data` instead of `Message`.

geiger/src/graph/Graph.js
```diff
@@ -0,0 +1,283 @@
...
+
+  /** @return {Array<Edge>} */
+  getRemovedEdges() {
+    return this._removed_edges;
+  }
+
+  // Cache the removed nodes/edges in the current batch, which will be used to
+  // remove nodes/edges from canvas.
+  // Start tracking changes for next batch.
+  beforeRender() {
```
hoch: Perhaps `PerformPreRenderTasks()`?

geiger/src/graph/Graph.js
```diff
@@ -0,0 +1,283 @@
...
+    this._removed_nodes = this._nodes.exit();
+    this._removed_edges = this._edges.exit();
+
+    // Clear the current tracked exit to track new changes in next batch.
+    this._nodes.clearEnterExit();
+    this._edges.clearEnterExit();
+  }
+
+  // After rendering the current batch, release the memory.
+  afterRender() {
```
hoch: Similarly, `PerformPostRenderTasks()`?

geiger/src/graph/NodeWithPort.js
```diff
@@ -0,0 +1,249 @@
...
+  /** @return {number} */
+  _getParamNumber() {
+    return this.getPortsByType(PortTypes.PARAM).length;
+  }
+
+  /**
+   * Use number of inputs and outputs to compute the layout
+   * for text and ports.
+   * @param {!NodeCreationMessage} message
+   * @credit This function is mostly borrowed from Audion/
```
hoch: `@credit`? Is this available somehow?

geiger/src/graph/Edge.js
gaopinghuang0: Done.

geiger/src/graph/Graph.js
gaopinghuang0: Done.

geiger/src/graph/Graph.js
gaopinghuang0: Done

geiger/src/graph/Graph.js
gaopinghuang0: Done.

PR #187, opened 2019-08-06
Implement Rootbeer ConstantSourceNode

rtoy:
Implements ConstantSourceNode with an AudioWorklet.  I think it's fully functional.

Used testharness and audit for testing, but maybe that's not what we want.  But it has a nice advantage that we could import WPT tests (with modifications) to test this node.

The actual Rootbeer ConstantSourceNode is defined in the test itself right now. I'll need to move that out somewhere.


hoch:
@rtoy Can you fix the check failure?

rootbeer/constant-source/test-constant-source.html
```diff
@@ -0,0 +1,183 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Rootbeer Test: ConstantSource</title>
+    <script src="../resources/testharness.js"></script>
+    <script src="../resources/testharnessreport.js"></script>
+    <script src="../resources/audit.js"></script>
```
hoch: I am not sure about this. Do we want these things in this project? Do you have a plan to integrate this project to Chromium at some point?

rootbeer/constant-source/test-constant-source.html
```diff
@@ -0,0 +1,183 @@
...
+  </head>
+
+  <body>
+    <script>
+      const audit = Audit.createTaskRunner();
+
+      let processorPath = "../processors/constant-source-processor.js";
```
hoch: Use `const` by default. Use `let` only when the value changes.

rootbeer/constant-source/test-constant-source.html
```diff
@@ -0,0 +1,183 @@
...
+
+      class RB_ConstantSourceNode extends AudioWorkletNode {
+        constructor(context, options) {
+          let procOptions = {
+            numberOfinputs : 0,
+            numberOfOutputs : 1,
+            parameterData : {offset : (options && options.offset) || 1}
+          };
+
+          console.log(`Node options`);
```
hoch: Why `console.log()`? Are we using this for testing?

hoch: Ditto for all `console.log()` instances below. I suggest to wrap this into an utility function so you can disable them easily.

rootbeer/processors/constant-source-processor.js
```diff
@@ -0,0 +1,86 @@
...
+  constructor(options) {
+    super(options);
+    // Start and stop times for the node.
+    this._startTime = Infinity;
+    this._stopTime = Infinity;
+    this._startFrame = Infinity;
+    this._stopFrame = Infinity;
+
+    this.port.onmessage = event => {
+      console.log(`Got processor event time = ${currentTime}`);
```
hoch: Ditto.

rootbeer/constant-source/test-constant-source.html
rtoy: As I mentioned, it's not necessary, but it makes it easier to pull in WPT tests to test this node.  Perhaps that's not important enough to have these, but writing tests take a long time and building off of existing tests would be useful.

hoch: That's a good point.

But then how do we pull these libraries when they are changed? (version update?) Also I don't see any script that handles dependencies. Manually copying files around is something that we don't want to do...

Also running a test seems very manual in the current form. Do you have a plan to run multiple tests with a script in the future?

hoch:
@rtoy I am not sure if we want to continue this effort. Looking back, this project is potentially huge, so it should be a separate repository. What do you think?

hoch:
We are not pursuing this anymore.

PR #196, opened 2020-05-17
fix ReferenceError of undefined object

### LINKED ISSUE #195: Blank Page Displays on Browsers that don't support Audio Worklet
Safari v13.0.4 shows a blank page when loading https://googlechromelabs.github.io/web-audio-samples/audio-worklet/

Console error:
```
[Error] ReferenceError: Can't find variable: OfflineAudioContext	
  _detectAudioWorklet (Components.js:210)	
  WorkletIndicator (Components.js:202)	
  build (Components.js:35)	
  Module Code (audio-worklet:52)	
  evaluate	
  moduleEvaluation
```

AnthumChris:
fixes detection on unsupported browsers (tested with Safari v13.0.4)

fixes #195

### DIFF
--- a/src/components/Components.js
+++ b/src/components/Components.js
@@ -207,9 +207,9 @@ const WorkletIndicator = () => {
 
 // Check if AudioWorklet is available.
 function _detectAudioWorklet() {
-  let context = new OfflineAudioContext(1, 1, 44100);
+  let context = window.OfflineAudioContext && new OfflineAudioContext(1, 1, 44100);
   return Boolean(
-      context.audioWorklet &&
+      context && context.audioWorklet &&
       typeof context.audioWorklet.addModule === 'function');
 }
 


AnthumChris:
Please ignore this PR.  I accidentally opened this on my forked repo, not the upstream.

src/components/Components.js
```diff
@@ -207,9 +207,9 @@ const WorkletIndicator = () => {
 
 // Check if AudioWorklet is available.
 function _detectAudioWorklet() {
-  let context = new OfflineAudioContext(1, 1, 44100);
+  let context = window.OfflineAudioContext && new OfflineAudioContext(1, 1, 44100);
```
hoch, line 210: 1. I have no idea about the current status of Safari. Did they remove "webkit" prefix?
2. Why do we have to check if OfflineAudioContext is valid? How does this help us detect Safari?

AnthumChris: This code doesn't necessarily fix safari, but it does prevent a TypeError that occurs when referencing a property/variable.  This would also fix detection of AudioWorklet on older browsers where the `OfflineAudioContect` property does not exist on window.

For example, `const a = FooBar` will throw a TypeError on all browsers, but `const a = window.FooBar` will not

hoch: I am confused. OfflineAudioContext exists on every browser. I think the clue that you want to check is the webkit prefix.

Thanks for the suggestion, but I prefer something more explicit and less magical. If something bad happens, it's better to throw, catch and display a helpful error message.

hoch:
@AnthumChris I believe now all the major browsers support the Audio Worklet. Is this PR still relevant?

AnthumChris:
@hoch I'm always a fan of showing unsupported messages for older browsers, but feel free to close this. Thanks for looking into it.

hoch:
Thanks for the response, @AnthumChris!

I'll try this PR locally and see how it works. Would love to learn how to detect Safari efficiently.

max-vogler:
I reproduced this in Safari 14. Although Safari [supports most features](https://github.com/chrisguttandin/standardized-audio-context) of OfflineAudioContext through `webkitOfflineAudioContext`, the website stays blank due to the ReferenceError. Merging this PR would improve the experience in Safari a lot!

image

max-vogler:
Ah, I realize that AudioWorklet is unavailable in Safari. A polyfill like `window.OfflineAudioContext ||= window.webkitOfflineAudioContext;` will not help. Nevertheless, we should still avoid the blank page to allow users of unsupported browsers to view the contents and examples.

src/components/Components.js
hoch: I hope we can expand l.210 with some comments. (like comments from @AnthumChris will be helpful when we read this code in the future)

cc/ @max-vogler

hoch:
I hope there is a better way to do this. Sniffing via OfflineAudioContext is an anti-pattern, and this was a quick-and-dirty check I came up with years ago.

max-vogler:
Detecting browser features by probing for the existence of a property on `window` is actually very idiomatic JavaScript. It is recommended by [MDN](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection) and used by libraries like [Modernizr](https://github.com/Modernizr/Modernizr/blob/master/feature-detects/audio/webaudio.js). Alternative approaches like user agent sniffing are very problematic, see [MDN: Browser detection](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent).

hoch:
Yes, I am aware of the common practice, but as an implementor creating a BaseAudioContext just for sniffing is rather somewhat expensive. I hope there's a better way.

AnthumChris:
We’re only checking for the existence of a global variable in Javascript scope. Let’s get this merged in so users don’t receive a blank page. User experience is important.

max-vogler:
I don't know the resource costs of instantiating an OfflineAudioContext. I only know that instantiating and probing for features in this way would give the highest confidence that the expected functionality is there. Especially if the current alternative is a blank page for a big amount of visitors. If @hoch would like to avoid the instantiation, here's an alternative:

```js
function _detectAudioWorklet() {
  return "OfflineAudioContext" in window && "AudioWorkletNode" in window;
}
```

This function takes the global existence of the `AudioWorkletNode` interface as proxy for the browser's AudioWorklet support. YMMV.

hoch:
A nice, short and no instantiation! This is a lot better than what I wrote. If I have to nit-pick...
- `"OfflineAudioContext" in window`: this will fail on WebKit. Right?
- `"AudioWorkletNode" in window`: this can also fail when the protocol is `http://`. (Secure Context)

Other than that - it looks good to me!

max-vogler:
`"OfflineAudioContext" in window` is `false` in Safari. I don't know of a browser that supports `AudioWorkletNode` only through a vendor-prefixed implementation like `webkitOfflineAudioContext`. We can either add a test for vendor-prefixed implementations or stop checking for OfflineAudioContext alltogether. Either:

```js
function _detectAudioWorklet() {
  return ("OfflineAudioContext" in window || "webkitOfflineAudioContext" in window) && "AudioWorkletNode" in window;
}
```

Or simply:
```js
function _detectAudioWorklet() {
  return "AudioWorkletNode" in window;
}
```

`"AudioWorkletNode" in window` does work for insecure (`http://`) contexts and returns `true`, see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorkletNode):
image.

AnthumChris:
also consider [`hasOwnProperty()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty) which has better backwards compat with older browsers (which reduces JavaScript syntax errors)

hoch:
Yup. This is my mistake. I used SecureContext on the spec, but didn't add that to Chromium IDL. :(

https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/webaudio/audio_worklet_node.idl
https://www.w3.org/TR/webaudio/#audioworkletnode

Thanks for finding this, @max-vogler! :)

AnthumChris:
@hoch @max-vogler Do we want to get this fixed and do we see value in fixing it?  At a minimum, it could encourage Safari's support for implementing new features that are only supported in other browsers.

Showing a blank page instead of an error does not seem to provide any advantage to anyone, especially the users/devs we're trying to advocate for:

image

max-vogler:
@hoch: Please pick one of the proposed solutions since you are the original reviewer of this PR. How about this one:

```js
function _detectAudioWorklet() {
  return (window.hasOwnProperty("OfflineAudioContext") || window.hasOwnProperty("webkitOfflineAudioContext")) && window.hasOwnProperty("AudioWorkletNode");
}
```

I am in favor of any, because all of the proposed solutions are a **big** improvement over having a blank page in a significant chunk of the browser market share.

max-vogler:
Thanks for fixing this, @AnthumChris!

AnthumChris:
I'm confused, was this deployed? Safari is still showing the error.

Screen Shot 2021-04-20 at 7 58 59 PM

hoch:
We serve pages directly from master. So I guess the PR merged above wasn't enough to fix the Safari problem. I'll take another look soon.

rtoy:
Most likely it's because `OfflineAudioContext` is undefined in Safari.  Safari still uses prefixed WebAudio.  The technology preview is where WebAudio becomes unprefixed.

PR #199, opened 2020-10-03
Fix confusing pointer math in HeapAudioBuffer

wrongbad:
Previously had an address right-shifting by the number of bytes in a uint16, which was non-sensical but working because right shifting by 2 is equivalent to dividing by the number of bytes in f32 (the actual intention). This change makes the code more faithful to intent.

### DIFF
--- a/audio-worklet/design-pattern/lib/wasm-audio-helper.js
+++ b/audio-worklet/design-pattern/lib/wasm-audio-helper.js
@@ -14,9 +14,6 @@
  * the License.
  */
 
-// Basic byte unit of WASM heap. (16 bit = 2 bytes)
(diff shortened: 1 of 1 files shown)


wrongbad:
@googlebot I signed it!

hoch (commented):
Sorry for the late response and thank you for your contribution.

So there's no functional change? Just clarifying pointer math?

wrongbad:
Yeah my intention was the values passed to `this._module.HEAPF32.subarray(` are exactly the same, but the computation of them is more intuitive/readable. It's been a long time since I looked at this so maybe a quick test to verify is warranted.

hoch (approved):

PR #204, opened 2021-02-14
Fix single quotes and trailing commas in json files

### LINKED ISSUE #193: JSON is not correct...
If these are supposed to be JSON files, they should have double quotes instead of single quotes and there should not be trailing commas.  It they are js files, I don't think you can {} on a line it would be assigned to a variable.  Cheers.

sukhbir192000:
Fixes #193


hoch (commented):
Thanks for your contribution!

This is already great, but could you make this truly compliant to JSON format? This is what I get from 01_Saw file:

```
Error: Parse error on line 2050:
...00000,		0.000000,	],	"imag": [		0.00
---------------------^
Expecting 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '[', got ']'
```

It seems like the trailing comma is causing this error.

sukhbir192000:
I fixed the commas and also tested it by parsing it to JSON. It's working now.

hoch:
Looks great and thank you!

Issue #205, opened 2021-04-07
Add audio recorder example


max-vogler:
- [x] Record, play, and remove audio tracks
- [x] Visualize audio with a simple waveform
- [x] Add README and inline documentation
- [x] Pause microphone access when no clip is being recorded
- [x] Save and load audio tracks
- [x] Reformat code following https://github.com/google/eslint-config-google
- [ ] Support downloading PCM, Opus
- [ ] Make it a PWA
  - [x] Icon, Manifest
  - [x] [Offline support](https://web.dev/offline-cookbook/)
  - [ ] [Share recorded audio](https://web.dev/web-share/)
- [ ] Color waveform based on audio
- [ ] Resize instead of scroll waveform?

https://googlechromelabs.github.io/web-audio-samples/audio-recorder/

max-vogler:
Closing, as the basic version of audio recorder has been finalized. I might revisit in the future to complete the stretch goals.

PR #207, opened 2021-04-22
Stop microphone access after recording is finished.

max-vogler:
See #205


hoch (commented):
Does this stop the stream from getUserMedia()? It looks like the code is stopping the media recorder.

How does the tab audio indicator look when you stop recording?

max-vogler:
This stops the stream from getUserMedia, which in turn stops the MediaRecorder and hides the tab's audio indicator.

hoch:
That's interesting. Thanks for the clarification, and this looks good to me.

Issue #214, opened 2021-04-30
Clean up and upgrade 2.1.0 


hoch:
- [x] Add https://github.com/google/eslint-config-google for automatic linting.
- [ ] ~~Upgrade dependencies.~~
- [ ] ~~Add a local dev server for easier development.~~
- [x] Revise the landing page so we have links to the Audio Recorder and the Drum Machine.

hoch:
Upgrading cause so many problems. It seems like gulp 4 and css plug-ins have vulnerability issues, and this makes the npm audit tool nags you forever. I am thinking of dropping gulp and plug-ins and switching to Eleventy + Tailwind.

```
url-regex  *
Severity: high
Regular Expression Denial of Service - https://npmjs.com/advisories/1550
fix available via `npm audit fix --force`
Will install gulp-concat-css@1.2.0, which is a breaking change
node_modules/url-regex
  rework-import  >=2.0.0
  Depends on vulnerable versions of url-regex
  node_modules/rework-import
    gulp-concat-css  >=2.0.0
    Depends on vulnerable versions of rework-import
    node_modules/gulp-concat-css

yargs-parser  <=13.1.1 || 14.0.0 - 15.0.0 || 16.0.0 - 18.1.1
Prototype Pollution - https://npmjs.com/advisories/1500
fix available via `npm audit fix --force`
Will install gulp@3.9.1, which is a breaking change
node_modules/yargs-parser
  yargs  4.0.0-alpha1 - 12.0.5 || 14.1.0 || 15.0.0 - 15.2.0
  Depends on vulnerable versions of yargs-parser
  node_modules/yargs
    gulp-cli  >=2.0.0
    Depends on vulnerable versions of yargs
    node_modules/gulp-cli
      gulp  >=4.0.0
      Depends on vulnerable versions of gulp-cli
      node_modules/gulp
```

hoch:
Just merged #218, #219, #220, #221, #222 to fix the last check box.

PR #215, opened 2021-05-06
Configure eslint.

max-vogler:
This PR adds eslint, configured per https://github.com/google/eslint-config-google and https://eslint.org/docs/user-guide/getting-started. You can run `npm run lint` to check the project. For now some folders like archive and audio-recorder are excluded. I'll format audio-recorder and remove it from the blocklist in a future PR.

On a side note: `package.json` is in `.gitignore` - is this intended? Typically the lockfile is checked into the source to pin versions and allow reproducible builds.

Related: #214

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,21 @@
+{
+    "env": {
+        "browser": true,
+        "es2021": true
+    },
+    "extends": [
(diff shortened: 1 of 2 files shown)


hoch:
Awesome! Thank you!

PR #225, opened 2021-06-04
Support offline mode through ServiceWorker.

max-vogler:
This PR is a child of #224 - please review the parent PR first. 

See #205 .


hoch (approved):
LGTM - thank you so much for working on this!

In a follow-up PR, can we do:
1. Adding Google license to each file.
2. Adding a top-level comment per file.

Issue #226, opened 2021-06-17
Revive Shiny Drum Machine


max-vogler:
Here are some ideas to revive Shiny Drum Machine:

- [x] Extract JavaScript to dedicated file.
- [x] Modernize HTML
  - [x] Migrate to HTML5
  - [x] Use Flex or Grid layouts
  - [x] Replace custom inputs with accessible, styleable HTML5 inputs (sliders, selects, buttons).
  - [ ] ~Mobile friendly responsive layout~
  - [x] Use `cursor: pointer` only for interactive elements
  - [x] Fix: Only trigger actions on left-click
- [x] Modernize JavaScript
- [ ] Add features
  - [ ] MIDI support from https://www.github.com/cwilso/MIDIDrums
  - [ ] Saving and loading from files

PR #227, opened 2021-06-17
Refactor Shiny Drum Machine to HTML5, Flex layout

max-vogler:
See #226.

 * Extracted JavaScript to dedicated file.
 * Refactored from HTML4 to HTML5.
 * Only use pointer cursor on clickable elements.
 * Refactored to flex layout, removed HTML spacing hacks.

The UI renders almost exactly the same. Notably, flexbox improves some vertical centerings: "Tempo" is now aligned with "100 bpm". "Demo" is now aligned with the numbered buttons.

image


hoch (approved):
I love this!

We don't have to do it in this PR, but perhaps we could modernize the JS file in a follow-up PR? (const, let, function notation etc.) Since you've already added a linter to this project, can we apply that to this project as well? :)

Just a suggestion. Not a requirement.

max-vogler:
I'll work a lot on the JavaScript in follow-up PRs, no worry ;)

Issue #232, opened 2021-07-15
Revive DJ sample code


max-vogler:
Sample: https://googlechromelabs.github.io/web-audio-samples/archive/demos/dj.html

- [x] Separate CSS, HTML, JavaScript
- [x] Clean up JavaScript to conform with modern ES features (fetch, classes)
- [x] Fix flaky AudioContext starting ("The AudioContext was not allowed to start. It must be resumed (or created) after a user gesture on the page. https://goo.gl/7K7WLu")
- [x] Evaluate porting the UI elements to https://github.com/nexus-js/ui/
- [x] Evaluate visualizing playback with waveform, oscilloscope, or similar
- [x] Use drag-and-drop to change played sample

PR #233, opened 2021-07-15
Modernize DJ sample, introduce Nexus.JS UI.

max-vogler:
Now:
image

Before:
image

* Extract JS and CSS from dj.html.
* Apply eslint to dj.js.
* Refactor to modern ES.
* Replace custom sliders with NexusUI dials and sliders.
* Replace custom time indicator with Nexus.JS dial.

We can discuss in our next sync: Nexus.JS has very nice controls but lacks in terms of accessibility and responsiveness.

See #232


max-vogler:
I added some visuals:

image

hoch (approved):
LGTM! The code looks a lot cleaner now!

PR #235, opened 2021-07-22
Redesign DJ sample to drag-n-drop based Material UI

max-vogler:
image

Redesigned DJ sample to drag-n-drop based Material UI. Cleaned up repetitive code by introducing a `DeckPlayer` class for both decks.

See #232

### DIFF
--- a/archive/demos/dj.js
+++ b/archive/demos/dj.js
@@ -4,10 +4,6 @@
 // init() once the page has finished loading.
 window.onload = init;
 
-let context;
-let timeIndicator;
-let timeIndicatorText;
-
 const tempo = 120.0; // hardcoded for now
 let anchorTime = 0;
 
@@ -25,7 +21,7 @@ const TRACKS = [
 
 const html = document.documentElement;
 
-async function fetchAndDecodeAudio(url) {
+async function fetchAndDecodeAudio(context, url) {
   const response = await fetch(url);
   const responseBuffer = await response.arrayBuffer();
 
@@ -68,18 +64,20 @@ async function loadTrackIntoDeck(deck, deckEl, trackId) {
   deck.loadTrack(getTrackURL(trackId));
 }
 
-function draw() {
+function draw(context, indicator, textEl) {
   // Calculate 4/4 beat position.
   const time = context.currentTime - anchorTime;
   const beat = (tempo / 60.0) * time;
   const roundedBeat = 4.0 * Math.floor(beat / 4.0);
   const wrappedBeat = beat - roundedBeat;
-  timeIndicator.value = wrappedBeat / 4;
+  indicator.value = wrappedBeat / 4;
 
   const fullBeat = 1 + Math.trunc(beat / 4);
-  timeIndicatorText.textContent = `${fullBeat}.${1 + Math.trunc(wrappedBeat)}`;
+  textEl.textContent = `${fullBeat}.${1 + Math.trunc(wrappedBeat)}`;
 
-  requestAnimationFrame(draw);
+  requestAnimationFrame(() => {
+    draw(context, indicator, textEl);
+  });
 }
 
 class DeckPlayer {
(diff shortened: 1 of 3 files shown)


archive/demos/dj.js
```diff
@@ -30,229 +31,263 @@ async function fetchAndDecodeAudio(url) {
...
-
-  // Create gain nodes to control the volume for the two sources.
-  source1Gain = context.createGain();
-  source2Gain = context.createGain();
+  // Create dynamics compressor to sweeten the overall mix.
+  const compressor = context.createDynamicsCompressor();
+  compressor.connect(postCompressorGain);
 
   // Create pre-compressor gain node.
   const preCompressorGain = context.createGain();
```
hoch, line 210: We should avoid the old factory pattern. We prefer:
```js
const preCompressorGain = new GainNode(context, {gain: 0.4});
```

hoch: Also ditto for all other factory methods.

archive/demos/dj.js
```diff
@@ -30,229 +31,263 @@ async function fetchAndDecodeAudio(url) {
...
-  // Create a lowpass resonant filter for both left and right
-  lowFilter1 = context.createBiquadFilter();
-  lowFilter2 = context.createBiquadFilter();
-  lowFilter1.frequency.value = 22050.0;
-  lowFilter2.frequency.value = 22050.0;
-  lowFilter1.Q.value = 5.0; // in decibels
-  lowFilter2.Q.value = 5.0; // in decibels
-
   // Create a convolver for a rhythm effect
   const convolver = context.createConvolver();
```
hoch, line 215: Ditto on the factory method.

hoch (approved):
LGTM with nits

archive/demos/dj.js
max-vogler: Thanks for the review! Could you explain why this is preferred? The [spec](https://webaudio.github.io/web-audio-api/#factory-method) seems unopinionated and [MDN](https://developer.mozilla.org/en-US/docs/Web/API/GainNode/GainNode) strongly prefers the factory methods.

hoch: I am not sure why MDN is documented that way, but it is wrong. We changed the spec because factory methods are generally bad, but especially for the extensibility. More info:
https://github.com/w3ctag/design-reviews/blob/main/2013/07/WebAudio.md#issue-constructibility--de-sugaring-of-static-methods

Also you can't use `GainNodeOptions` object with the factory method.

max-vogler: Interesting! I'll follow up with MDN and change the code accordingly.

max-vogler: Ah, I guess Safari compatibility might be a reason. Safari only supports the constructor since April 2021.

[new GainNode()](https://caniuse.com/mdn-api_gainnode_gainnode) has 84.4 % support
[createGain()](https://caniuse.com/mdn-api_baseaudiocontext_creategain) has 95.2 % support

hoch:
```
const context = new AudioContext();
```

I am quite certain the 15.6% of Safari won't pass the line above for the same reason. Have you tested the app on the most recent iOS?

max-vogler: Fair point, the unprefixed AudioContext constructor has the same support in Safari. I'm fine with either changing the code back to increase compatibility, or changing it forward to show the modern API. Which SLO for browser compatibility do you suggest @hoch?

max-vogler: As a side note: I test the web app on Safari 14.1.1 on macOS 11.4 and it works.

max-vogler: Raised https://github.com/mdn/content/issues/7426 with MDN.

hoch: Thanks! If its most recent version supports the modern Web Audio correctly, then we don't have any blocker! :)

max-vogler: Done.

PR #252, opened 2022-08-08
Introduce Worklet Recording App

eldrickb:
This PR introduces a recording demo using AudioWorklet. Visually and functionally, it's a mirror of the ScriptProcessorNode recording demo.  

Basic functions:
-   Records audio using Audio Worklet
-   Presents the recording for playback or download
-   Renders a gain visualizer for live and recorded audio

Some notes: 
- The exporter module doesn't write a proper audio file. Length, file properties, etc are written correctly, but audio playback sounds blank. 
	- Every other aspect of the app is done, so knowing that this is your code and you're the expert in this area, I was hoping to delegate investigation of this bug to you as I work on more important tasks. We can definitely chat about this tomorrow (Monday)
- As far as I know there is no way to create a "shared audio buffer", so the processor interleaves the audio data onto a shared array buffer live.
- Some things I did in this app might not be the most optimal / elegant (e.g. the way setupRecordingWorkletNode is written), and as I realize we write these demos to be read and dissected, I wasn't sure if it might seem a little convoluted. Your thoughts?


src/audio-worklet/migration/worklet-recorder/app.js
```diff
@@ -0,0 +1,339 @@
...
+ *    Function to call to trigger recording state updates in the processor.
+ */
+
+/**
+ * Creates ScriptProcessor to record and track microphone audio.
+ * @param {number} maxLength Maximum recording length, in seconds.
+ * @return {object} Object containing return components.
+ * @return {RecordingComponents} Recording node related components for the app.
+ */
+async function setupRecordingWorkletNode(
```
hoch, line 93: The line wrapping looks a bit strange. Can you check?

src/audio-worklet/migration/worklet-recorder/app.js
```diff
@@ -0,0 +1,339 @@
...
+    maxLength,
+) {
+  // A float is 4 bytes, so our length must be 4x
+  const recordingBuffer = new Float32Array(
+      new SharedArrayBuffer(context.sampleRate * maxLength * 4));
```
hoch, line 98: So I assume this doesn't work from a domain that does not support COEP/COOP?

hoch (commented):
A high-level comment: can we build the app without SharedArrayBuffer?

eldrickb:
Ready for review.

hoch (approved):
LGTM

PR #279, opened 2023-03-30
Trying out WGSL for doing computations in GPU

Dp-Goog:
Use WebGPU shader language (WGSL) to write shaders that perform computations. The comments in the code should pinpoint the path taken to go from the writing all the way to the reading step.

The WORKGROUP_SIZE variable is used to parallelize the operation on WebGPU.


hoch (approved):
LGTM. Thanks for digging into this!

PR #281, opened 2023-03-30
Implementing convolution using WGSL on WebGPU

Dp-Goog:
v1: Trying out convolution using WebGPU. Still looking into how to parallelize this work.
v2: Parallelize convolution by dispatching in work that is the size of the convolution array. Added a simple test that verifies that the IrHelper generated TestIR works properly by passing in an impulse pulse as an input.


src/experiments/webgpuaudio/main.js
```diff
@@ -13,6 +13,10 @@ let toggleButton = null;
 let audioContext = null;
 let isPlaying = false;
 
+// Change this value to true to run tests and see test errors on JS console
+// during audio processing. Not intended for production, only for testing.
+let runTests = false;
```
hoch, line 18: How about having `TEST_MODE = false` in `constants.js` and use it by importing the file everywhere?

hoch: Then we don't have to pass it around via postMessage and alike.

src/experiments/webgpuaudio/test/test_processor.js
```diff
@@ -0,0 +1,28 @@
+import GPUProcessor from "../gpu-processor.js";
+import IRHelper from "../ir-helper.js"
+
+class TestProcessor {
+
+    constructor() {}
+    testConvolution = async() => {
```
hoch, line 7: I was surprised that this syntax works, but the formal ES6 notation would be:
```js
constructor() {}

async testConvolution() {
...
}
```

src/experiments/webgpuaudio/main.js
Dp-Goog: Good idea, done!

src/experiments/webgpuaudio/test/test_processor.js
Dp-Goog: Done!

hoch (approved):
LGTM

PR #282, opened 2023-03-30
Adding ir-helper.js and refactor worker.js

hoch:
- Added `ir-helper.js` for a simple IR buffer (Float32Array) creation. 
- Refactor `worker.js` for better readability and modularity.

### DIFF
--- a/src/experiments/webgpuaudio/worker.js
+++ b/src/experiments/webgpuaudio/worker.js
@@ -1,5 +1,6 @@
 import FreeQueue from "./lib/free-queue.js";
 import GPUProcessor from "./gpu-processor.js";
+import IRHelper from "./ir-helper.js"
 import { FRAME_SIZE } from "./constants.js";
 
 /**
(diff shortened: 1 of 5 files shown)


Dp-Goog:
Looks good so far!

hoch:
I'll do below with a follow-up PR:

- [ ] Add a function that accepts AudioBuffer and extract an 1D Float32Array for the IR usage.

src/experiments/webgpuaudio/worker.js
```diff
@@ -1,44 +1,71 @@
...
-    }
+// Harmful globals
+let inputQueue = null;
+let outputQueue = null;
+let atomicState = null;
+let gpuProcessor = null;
+let inputBuffer = null;
+
+// Performance metrics
+let lastCallback = 0;
```
Dp-Goog, line 14: Should this be initialized to performance.now() instead of 0?

Dp-Goog: Sounds good, thanks for explaining!

Dp-Goog (approved):

PR #284, opened 2023-03-30
Add audio file utilities and fix other files that use ir-helper.js

hoch:


src/experiments/webgpuaudio/ir-helper.js
```diff
@@ -25,6 +25,30 @@ const createTestIR = (length) => {
...
+ * @param {string} url A URL to the IR audio file.
+ * @returns {Float32Array}
+ * 
+ * @example
+ * const irArray = await fetchAudioFileToF32Array(
+ *     audioContext,
+ *     '../../sounds/impulse-responses/cardiod-35-10-spread.wav');
+ * console.log(array);    // You'll see a big array!!! 
+ */
+const fetchAudioFileToF32Array = async (audioContext, url) => {
```
Dp-Goog, line 43: nit: I wonder if we should just rename this file to helper.js, or audio-helper.js, since this is also doing more than just returning the impulse response. What do you think?

Dp-Goog (approved):
LGTM, thanks for fixing the lint/styles of the test as well!

src/experiments/webgpuaudio/ir-helper.js
hoch: Fair point, but currently both functions are for creating or loading IR files. If the scope expands in the future, I'll definitely change the name.

Issue #292, opened 2023-04-10
"AudioWorklet, SharedArrayBuffer, and Worker" demo link broken


TenghuiZhang:
The demo in the [AudioWorklet, SharedArrayBuffer, and Worker](https://googlechromelabs.github.io/web-audio-samples/audio-worklet/design-pattern/shared-buffer/) page is not function. 

The error log is 
`[SharedBufferWorklet] Worker Error: SharedArrayBuffer is not supported in your browser. See
          https://developers.google.com/web/updates/2018/06/audio-worklet-design-pattern
          for more info.`

TenghuiZhang:
Based on the new SharedArrayBuffer Chrome [security requirement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements),  we need to add the two headers need to be set to cross-origin isolate site. 
Header: 
`Cross-Origin-Opener-Policy: same-origin
Cross-Origin-Embedder-Policy: require-corp
`
 
The useful tool to implement are [ServiceWorker COOP/COEP](https://github.com/gzuidhof/coi-serviceworker)

PR #293, opened 2023-04-10
Fixing the "AudioWorklet, SharedArrayBuffer, and Worker" demo broken …

TenghuiZhang:
#292 Broken link fixing solution.

- Fixing the broken demo in https://googlechromelabs.github.io/web-audio-samples/audio-worklet/design-pattern/shared-buffer/


hoch:
I would recommend one of these solutions:
1. Create `src/third_party/` and add `coi-serviceworker.js` to use it in this repo.
2. Use an external CDN and access it from the page directly.

Issue #298, opened 2023-04-21
Broken download file button in "AudioWorklet Recorder"


TenghuiZhang:
Reproducing steps:
1. Navigate to https://googlechromelabs.github.io/web-audio-samples/audio-worklet/migration/worklet-recorder/
2. Start recording in any length
3. Strop recording
4. Click the download button
5. The download file always 4KB.

PR #301, opened 2023-04-24
Fixing the AudioWorklet Recorder download button link

TenghuiZhang:
Fixing #298 The download button is not functional

### DIFF
--- a/src/audio-worklet/migration/worklet-recorder/app.js
+++ b/src/audio-worklet/migration/worklet-recorder/app.js
@@ -327,7 +327,7 @@ const createRecord = (recordingProperties, recordingLength, sampleRate,
   const recordingUrl = createLinkFromAudioBuffer(recordingBuffer, true);
 
   player.src = recordingUrl;
-  downloadLink.src = recordingUrl;
+  downloadLink.href = recordingUrl;
   downloadLink.download =
     `recording-${new Date().getMilliseconds().toString()}.wav`;
   downloadButton.disabled = false;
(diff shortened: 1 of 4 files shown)


src/audio-worklet/migration/worklet-recorder/app.js
```diff
@@ -327,7 +327,7 @@ const createRecord = (recordingProperties, recordingLength, sampleRate,
   const recordingUrl = createLinkFromAudioBuffer(recordingBuffer, true);
 
   player.src = recordingUrl;
-  downloadLink.src = recordingUrl;
+  downloadLink.href = recordingUrl;
```
hoch, line 330: Hah! That's a nice catch. Do you have a staging page link from your fork?

TenghuiZhang: No, I didn't. And I am wondering why "Build and deploy to GitHub Pages / Build and deploy to GitHub Pages (pull_request)" It seems like I can not pull request to the main branch due to the deploy key?

hoch: If the auto deploy is not possible, could you locally check if the patch fixes the issue? (I believe you already did, but just to double check)

TenghuiZhang: Yes, the issue is fixed

TenghuiZhang:
The github pre-checking block is passed with the newest permission setup

PR #260, opened 2022-11-10
Fix functionality of "Turn monitor on/off" button in AudioWorklet Recorder

kk-kd:
https://googlechromelabs.github.io/web-audio-samples/audio-worklet/migration/worklet-recorder/

Right now the "Turn monitor on/off" button on the page isn't controlling monitor behavior. Proposing this patch, but happy to make modifications if you want another behavior.

PR #309, opened 2023-04-26
 Fix functionality of "Turn monitor on/off" button in AudioWorklet Recorder

kk-kd:
#260

### DIFF
--- a/src/audio-worklet/migration/worklet-recorder/app.js
+++ b/src/audio-worklet/migration/worklet-recorder/app.js
@@ -48,7 +48,7 @@ async function init() {
 
   // We can pass this port across the app
   // and let components handle their relevant messages
-  const visualizerCallback = setupVisualizers(recordingNode);
+  const visualizerCallback = setupVisualizers(monitorNode);
   const recordingCallback = handleRecording(
       recordingNode.port, recordingProperties);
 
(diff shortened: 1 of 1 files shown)


kk-kd:
It will probably be the easiest if I redo this commit... Let me redo it today after work.

TenghuiZhang:
Thank you so much for helping fixing this issue!

TenghuiZhang:
Hi, @kk-kd I just copy your code and run locally. There is a continuous and loud noise happen when the "Turn monitor" is ON (No recording action). Could you please help me verify is that happen on your side too?

hoch:
@TenghuiZhang Since you reviewed this change - could you explain how this change fixes the problem? Also what was the root cause of the issue? I'd be great if we can have that information in the description of the PR. (Or you can simply write it down as a comment.)

TenghuiZhang:
Yeah! This change use the "gain.value" of monitor node to control the visualization graph amplitude. When the minor button is off, the gain.value of monitorNode is 0. When the minor button is on, the gain value will change to 1. Therefore, we are able to control the amplitude of the graph by multiply monitorNode.gain.value. By using this strategy, we can control monitor behavior.

hoch (approved):
LGTM

PR #311, opened 2023-04-27
Fixing and improve the SPN recording demo

### LINKED ISSUE #308: Fixing and improve User experience of SPN Recorder demo
The max record length is 10s, but the recording can go infinitely. It should be auto paused. 
Since we agree to expend the max recording length to 5 mins, so we need to update this demo too. 
Edit the description of this demo to make the user understand what's is this demo stands for.

TenghuiZhang:
Fix #308 
Now the SPN recording demo is behave the same as Worklet recording demo
This PR also fix the format issue of github config file(main.yml).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out this repo
-      uses: actions/checkout@v3
+      uses: actions/checkout@v2
     - name: Install Node.js
       uses: actions/setup-node@v3
       with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out this repo
-      uses: actions/checkout@v2
+      uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675
     - name: Install Node.js
       uses: actions/setup-node@v3
       with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out this repo
-      uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675
+      uses: actions/checkout@v3
     - name: Install Node.js
       uses: actions/setup-node@v3
       with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out this repo
-        uses: actions/checkout@v2
+      uses: actions/checkout@v2
     - name: Install Node.js
       uses: actions/setup-node@v3
       with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out this repo
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
     - name: Install Node.js
       uses: actions/setup-node@v3
       with:
--- a/src/audio-worklet/migration/spn-recorder/app.js
+++ b/src/audio-worklet/migration/spn-recorder/app.js
@@ -42,7 +42,7 @@ async function init() {
       echoCancellation: false,
       autoGainControl: false,
       noiseSuppression: false,
-      latency: 0,
+      latency: 0
     },
   });
 
@@ -52,7 +52,7 @@ async function init() {
   const recordingProperties = {
     numberOfChannels: 2,
     sampleRate: context.sampleRate,
-    maxFrameCount: context.sampleRate * 300,
+    maxFrameCount: context.sampleRate * 300
   };
 
   // Obtain samples passthrough function for visualizers
--- a/src/audio-worklet/migration/spn-recorder/app.js
+++ b/src/audio-worklet/migration/spn-recorder/app.js
@@ -11,9 +11,16 @@ const context = new AudioContext();
 // Arbitrary buffer size, not specific for a reason
 const BUFFER_SIZE = 256;
 let recordingLength = 0;
+let recordBuffer = [[], []];
 let isRecording = false;
 let isMonitoring = false;
 let visualizationEnabled = true;
+
+let recordButton = document.querySelector('#record');
+let recordText = recordButton.querySelector('span');
+let player = document.querySelector('#player');
+let downloadButton = document.querySelector('#download-button');
+let downloadLink = document.querySelector('#download-link');
 
 // Wait for user interaction to initialize audio, as per specification.
 document.addEventListener('click', (element) => {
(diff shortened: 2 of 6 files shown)


src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -11,10 +11,17 @@ const context = new AudioContext();
 // Arbitrary buffer size, not specific for a reason
 const BUFFER_SIZE = 256;
 let recordingLength = 0;
+let recordBuffer = [[], []];
 let isRecording = false;
 let isMonitoring = false;
 let visualizationEnabled = true;
 
+let recordButton = document.querySelector('#record');
+let recordText = recordButton.querySelector('span');
```
hoch, line 20: I recommend using `id="record-text"` instead of selecting `span`. There could be others.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -90,8 +101,8 @@ function setupScriptProcessor(recordBuffer, passSampleToVisualizers) {
 
       // While recording, feed data to recording buffer at the proper time.
       if (isRecording) {
-        recordBuffer
-            .copyToChannel(currentSamples[channel], channel, recordingLength);
+        let frameNumber = recordingLength / BUFFER_SIZE;
+        recordBuffer[channel][frameNumber] = currentSamples[channel].slice();
```
hoch, line 105: What does this do?

Also `.slice()` creates a new instance. Is this intended?

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -101,6 +112,21 @@ function setupScriptProcessor(recordBuffer, passSampleToVisualizers) {
     // Update tracked recording length.
     if (isRecording) {
       recordingLength += BUFFER_SIZE;
+      
+      if (recordingLength > recordingProperties.maxFrameCount) {
+        isRecording = !isRecording;
+
+        const finalRecordBuffer =
+            createFinalRecordBuffer(recordingProperties);
+        const wavUrl = createLinkFromAudioBuffer(finalRecordBuffer, true);
```
hoch, line 121: Let's rename `wavUrl` to `audioFileUrl`.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -101,6 +112,21 @@ function setupScriptProcessor(recordBuffer, passSampleToVisualizers) {
...
+        player.src = wavUrl;
+        downloadLink.href = wavUrl;
+        downloadLink.download =
+            `recording-${new Date().getMilliseconds().toString()}.wav`;
+        downloadButton.disabled = false;
+        recordText.innerHTML = 'Ready to download 5 mins';
```
hoch, line 127: `innerHTML` to `textContent`.

Also are we downloading a "5-minute file" always?

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -305,3 +325,25 @@ function setupRecordingGainVis() {
...
 }
+
+/**
+ * Create the recording buffer with right size
+ * @param {object} recordingProperties Properties of record buffer
+ * @return {AudioBuffer} Record buffer for current recording
+ */
+const createFinalRecordBuffer = (recordingProperties) => {
+  const contextRecordBuffer = context.createBuffer(
+      2, recordingLength, context.sampleRate,);
```
hoch, line 336: One redundant comma at the end.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -305,3 +325,25 @@ function setupRecordingGainVis() {
...
+  let startPos = 0;
```
hoch, line 337: Avoid abbreviation if possible: `startPosition`.

Also "position" is rather abstract. Is this an index of an array? Then say `startIndex` and add comments for it.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -305,3 +325,25 @@ function setupRecordingGainVis() {
...
+
+  for (let frame = 0; frame < recordBuffer[0].length; frame++){
+    for (let channel = 0; 
+        channel < recordingProperties.numberOfChannels; channel++) {
+
```
hoch, line 342: A blank line here doesn't seem necessary.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -305,3 +325,25 @@ function setupRecordingGainVis() {
...
+ * @return {AudioBuffer} Record buffer for current recording
+ */
+const createFinalRecordBuffer = (recordingProperties) => {
+  const contextRecordBuffer = context.createBuffer(
+      2, recordingLength, context.sampleRate,);
+  let startPos = 0;
+
+  for (let frame = 0; frame < recordBuffer[0].length; frame++){
+    for (let channel = 0; 
+        channel < recordingProperties.numberOfChannels; channel++) {
```
hoch, line 341: I'd do

```js
for (let channel = 0;
     channel < recordingProperties.numberOfChannels;
     channel++) {
```

hoch (commented):
Getting there! I made some comments.

src/audio-worklet/migration/spn-recorder/app.js
TenghuiZhang: "=" in js make Object, array a shallow copy. We have to use slice to create a deep copy. Therefore, the data in recordingBuffer will be modified based on currentSamples[channel] update. 

There is another way is copy the float number in array one by one since "=" in js make number a deep copy. Due to the time and readability consuming, I think using 'slice' is better.

src/audio-worklet/migration/spn-recorder/app.js
TenghuiZhang: No, the newest algorithm can make the download always be the right length.

src/audio-worklet/migration/spn-recorder/app.js
hoch: Understood, but since this is happening around Float32Array, I recommend using `.set()` method for faster operation: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set

It might need a bit or writing, but could you give some thoughts?

src/audio-worklet/migration/spn-recorder/app.js
hoch: Then we need to change the text to reflect what's happening.

src/audio-worklet/migration/spn-recorder/app.js
TenghuiZhang: Based on the reference, we only need to assign a new Float32Array. 

recordBuffer[channel][frameNumber] = new Float32Array(currentSamples[channel]);

In this way, we don't need set() OR slice(). I think these two way speed are similar. Basically, we are creating a new float32array.

.github/workflows/main.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out this repo
-      uses: actions/checkout@v2
+      uses: actions/checkout@a81bbbf8298c0fa03ea29cdc473d45769f953675
```
hoch, line 18: This doesn't seem relevant to this change. Can you explain?

src/audio-worklet/migration/worklet-recorder/app.js
```diff
@@ -31,17 +31,15 @@ async function init() {
...
   });
 
   const micSourceNode = context.createMediaStreamSource(micStream);
 
   const recordingProperties = {
     numberOfChannels: micSourceNode.channelCount,
     sampleRate: context.sampleRate,
-    maxFrameCount: context.sampleRate * 300,
-  };
+    maxFrameCount: context.sampleRate * 300};
```
hoch, line 42: Let's move the ending curly brace to the next line.

hoch (approved):
LGTM with nits

.github/workflows/main.yml
TenghuiZhang: I am still in the process of figuring it out. The reason I am doing it is to pass the github check. We have this issue before, but at that time, I pass the test by changing the checkout@v3 to checkout@v2 and add the permission:write. This time, the presubmit is blocked again. Haven't find a right solution yet.

PR #312, opened 2023-04-27
Fixing the SPN broken demo and formating the worklet demo

TenghuiZhang:
Fixing #308

### DIFF
--- a/src/audio-worklet/migration/spn-recorder/app.js
+++ b/src/audio-worklet/migration/spn-recorder/app.js
@@ -11,9 +11,16 @@ const context = new AudioContext();
 // Arbitrary buffer size, not specific for a reason
 const BUFFER_SIZE = 256;
 let recordingLength = 0;
+let recordBuffer = [[], []];
 let isRecording = false;
 let isMonitoring = false;
 let visualizationEnabled = true;
+
+let recordButton = document.querySelector('#record');
+let recordText = document.querySelector('#record-text');
+let player = document.querySelector('#player');
+let downloadButton = document.querySelector('#download-button');
+let downloadLink = document.querySelector('#download-link');
 
 // Wait for user interaction to initialize audio, as per specification.
 document.addEventListener('click', (element) => {
@@ -35,30 +42,30 @@ async function init() {
       echoCancellation: false,
       autoGainControl: false,
       noiseSuppression: false,
-      latency: 0,
+      latency: 0
     },
   });
 
   const micSourceNode = context.createMediaStreamSource(micStream);
 
-  // Prepare buffer for recording.
-  const recordBuffer = context.createBuffer(
-      2,
-      // 10 seconds seems reasonable for demo purposes.
-      context.sampleRate * 10,
-      context.sampleRate,
-  );
+  // Prepare max recording buffer for recording.
+  const recordingProperties = {
+    numberOfChannels: 2,
+    sampleRate: context.sampleRate,
+    maxFrameCount: context.sampleRate * 300
+  };
 
   // Obtain samples passthrough function for visualizers
   const passSampleToVisualizers = setupVisualizers();
-  const spNode = setupScriptProcessor(recordBuffer, passSampleToVisualizers);
+  const spNode =
+      setupScriptProcessor(recordingProperties, passSampleToVisualizers);
 
   const monitorNode = context.createGain();
   const inputGain = context.createGain();
   const medianEnd = context.createGain();
 
   setupMonitor(monitorNode);
-  setupRecording(recordBuffer);
+  setupRecording(recordingProperties);
 
   micSourceNode
       .connect(inputGain)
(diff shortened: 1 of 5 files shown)


src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -35,30 +42,30 @@ async function init() {
...
-      // 10 seconds seems reasonable for demo purposes.
-      context.sampleRate * 10,
-      context.sampleRate,
-  );
+  // Prepare max recording buffer for recording.
+  const recordingProperties = {
+    numberOfChannels: 2,
+    sampleRate: context.sampleRate,
+    maxFrameCount: context.sampleRate * 300
+  };
```
hoch, line 56: This is a lot easier to read! 👍

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -70,17 +77,21 @@ async function init() {
...
+function setupScriptProcessor(recordingProperties, passSampleToVisualizers) {
   const processor = context.createScriptProcessor(BUFFER_SIZE);
-  const currentSamples = new Array(recordBuffer.numberOfChannels).fill([]);
+  const currentSamples =
+     new Array(recordingProperties.numberOfChannels).fill([]);
 
   // Main SPN callback. Handles recording data and tracking recording length.
   processor.onaudioprocess = function(event) {
+    // Display current recording length.
+    document.querySelector('#data-len').innerHTML =
```
hoch, line 93: `innerHTML` to `textContent`

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -70,17 +77,21 @@ async function init() {
...
+    Math.round(recordingLength / recordingProperties.sampleRate * 100)/100;
```
hoch, line 94: This line needs more indentation?

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -90,8 +101,9 @@ function setupScriptProcessor(recordBuffer, passSampleToVisualizers) {
 
       // While recording, feed data to recording buffer at the proper time.
       if (isRecording) {
-        recordBuffer
-            .copyToChannel(currentSamples[channel], channel, recordingLength);
+        let frameNumber = recordingLength / BUFFER_SIZE;
```
hoch, line 104: With this operation, `frameNumber` can be a float. Is that okay?

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -111,37 +138,31 @@ function setupScriptProcessor(recordBuffer, passSampleToVisualizers) {
...
  */
-function setupRecording(recordBuffer) {
-  const recordButton = document.querySelector('#record');
-  const recordText = recordButton.querySelector('span');
-  const player = document.querySelector('#player');
-  const downloadButton = document.querySelector('#download');
-
-  async function prepareClip() {
+function setupRecording(recordingProperties) {
+  async function prepareClip(finalRecordBuffer) {
```
hoch, line 144: Can we take this function out of `setupRecording` and add comments?

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -305,3 +326,26 @@ function setupRecordingGainVis() {
 
   return draw;
 }
+
+/**
+ * Create the recording buffer with right size
+ * @param {object} recordingProperties Properties of record buffer
+ * @return {AudioBuffer} Record buffer for current recording
```
hoch, line 333: `@returns`

https://jsdoc.app/tags-returns.html

hoch (commented):
Thanks for your patience! We're almost there.

src/audio-worklet/migration/spn-recorder/app.js
TenghuiZhang: JS don't have the different number type like int, float, or double. The only type is number. 

Based on my design, the recordingLength is increased by BUFFER_SIZE each time (line 115). Therefore, recordingLength/BUFFER_SIZE will always be an "INTEGER"

src/audio-worklet/migration/spn-recorder/app.js
TenghuiZhang: Yes, it is moved.

src/audio-worklet/migration/spn-recorder/app.js
hoch: > JS don't have the different number type like int, float, or double. The only type is number.

Yes. I am well aware of that. (and that can be a "bad part" for JS audio programming) The reason why I am asking this is that when this `frameNumber` becomes a floating point number it creates a confusion when we use this to pinpoint a sample value in the array. 

```js
a = [1, 2, 3]
a[0.75] // undefined
```

The resulting value of `undefined` will create `NaN` values in the audio stream for subsequent operation. (That's very bad.) So it is a general practice in audio programming to make it clear that the sample index number is a positive integer.

> Based on my design, the recordingLength is increased by BUFFER_SIZE each time (line 115). Therefore, recordingLength/BUFFER_SIZE will always be an "INTEGER"

You have two options:
1) Add a comment that explains why this operation is safe for sample indexing.
2) Use `Math.floor()` to make sure it falls on to an integer.

Does it make sense?

hoch (approved):
LGTM!

PR #317, opened 2023-05-17
Splitting the visualization of canvas and improve readability of visulization

### LINKED ISSUE #316: AudioWorklet Recorder doesn't display waveform correctly.
1. Go to https://googlechromelabs.github.io/web-audio-samples/audio-worklet/migration/worklet-recorder/
2. Click the page and start recording.
3. Nothing is displayed in the Recording waveform visualization.
4. However, playing the recorded audio works fine.

The screenshot after recording ~5 seconds of audio:
image

TenghuiZhang:
Fix #316 

Splitting the gain value which is used in each visualization canvas.
Multiple the amplitude with a constant number 5 to improve the visualization.

### DIFF
--- a/src/audio-worklet/migration/spn-recorder/app.js
+++ b/src/audio-worklet/migration/spn-recorder/app.js
@@ -10,6 +10,8 @@ const context = new AudioContext();
 
 // Arbitrary buffer size, not specific for a reason
 const BUFFER_SIZE = 256;
+// Make the visulization more clear to the users
+const WAVEFROM_SCALE_FACTOR = 5
 let recordingLength = 0;
 let recordBuffer = [[], []];
 let isRecording = false;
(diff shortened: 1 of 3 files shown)


src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -237,10 +238,14 @@ function setupVisualizers() {
 
       currentSampleGain /= (currentSamples.length * currentSamples[0].length);
 
-      drawLiveGain(currentSampleGain);
+      const liveGain = currentSampleGain * monitorNode.gain.value;
+      // "Times 5" make the visulization more clear to the users
+      drawLiveGain(liveGain * 5);
```
hoch, line 243: I would create `WAVEFROM_SCALE_FACTOR = 5` at the top of this file with a relevant comment.

Do you know the max/min value of `liveGain`? That helps you decide the scale factor accordingly.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -237,10 +238,14 @@ function setupVisualizers() {
...
 
       if (isRecording) {
-        drawRecordingGain(currentSampleGain);
+        const recordGain = currentSampleGain;
+        // "Times 5" make the visulization more clear to the users
+        drawRecordingGain(recordGain * 5);
```
hoch, line 248: Ditto.

hoch (commented):
Let's think about the right approach of using a scale factor.

src/audio-worklet/migration/spn-recorder/app.js
TenghuiZhang: That's a good point! The canvas height is 150, so I believe the max value of Y is 75. Actually, I choose 5 is based on the manual testing visualization. Based on the calculation, I can not find a good factor. Do you know any formula which might help us to identify a better value for WAVEFROM_SCALE_FACTOR?

hoch (approved):
LGTM

hoch (approved):
LGTM

PR #331, opened 2023-05-25
Improve the sound wave of recording canvas

### LINKED ISSUE #330: Enlarge the waveform display of the Recording section.
Improve sound wave UI for better readability

TenghuiZhang:
This is the PR to fix #330

### DIFF
--- a/src/audio-worklet/migration/spn-recorder/app.js
+++ b/src/audio-worklet/migration/spn-recorder/app.js
@@ -7,10 +7,10 @@
 import createLinkFromAudioBuffer from './exporter.mjs';
 
 // This enum states the current recording state
-const recordingStates = {
-  notInitialized: 'notInitialized',
-  startIniAndRecording: 'startIniAndRecording',
-  finishRecording: 'finishRecording'
+const recorderState = {
+  UNINITIALIZED: 0,
+  RECORDING: 1,
+  FINISHED: 2,
 };
 
 const context = new AudioContext();
@@ -22,7 +22,7 @@ const WAVEFROM_SCALE_FACTOR = 5
 let recordingLength = 0;
 let recordBuffer = [[], []];
 let isRecording = false;
-let recordingState = recordingStates.notInitialized;
+let recordingState = recorderState.UNINITIALIZED;
 
 let recordButton = document.querySelector('#record');
 let recordText = document.querySelector('#record-text');
@@ -31,17 +31,15 @@ let player = document.querySelector('#player');
 let downloadButton = document.querySelector('#download-button');
 let downloadLink = document.querySelector('#download-link');
 
-
-
 // Wait for user interaction to initialize audio, as per specification.
-if (recordingState === recordingStates.notInitialized){
+if (recordingState === recorderState.UNINITIALIZED) {
   recordButton.disabled = false;
   recordButton.addEventListener('click', (element) => {
     init();
     isRecording = true;
-    recordingState = recordingStates.startIniAndRecording;
+    recordingState = recorderState.RECORDING;
     recordText.textContent = 'Continue';
-    changeButtonStatusIfNeeded();
+    changeButtonStatus();
   }, {once: true});
 }
 
@@ -69,7 +67,7 @@ async function init() {
   const recordingProperties = {
     numberOfChannels: 2,
     sampleRate: context.sampleRate,
-    maxFrameCount: context.sampleRate * 5
+    maxFrameCount: context.sampleRate * 300
   };
 
   const gainNode = context.createGain();
@@ -132,7 +130,7 @@ function setupScriptProcessor(recordingProperties, passSampleToVisualizers) {
       recordingLength += BUFFER_SIZE;
       if (recordingLength > recordingProperties.maxFrameCount) {
         isRecording = !isRecording;
-        recordingState = recordingStates.finishRecording;
+        recordingState = recorderState.FINISHED;
         recordText.textContent = 'Reach the maximum length of';
         const finalRecordBuffer =
             createFinalRecordBuffer(recordingProperties);
@@ -159,7 +157,7 @@ function setupScriptProcessor(recordingProperties, passSampleToVisualizers) {
 function setupRecording(recordingProperties) {
   recordButton.addEventListener('click', (event) => {
     isRecording = true;
-    changeButtonStatusIfNeeded();
+    changeButtonStatus();
   });
 
   stopButton.addEventListener('click', (event) => {
@@ -167,13 +165,13 @@ function setupRecording(recordingProperties) {
     isRecording = false;
     const finalRecordBuffer = createFinalRecordBuffer(recordingProperties);
     prepareClip(finalRecordBuffer);
-    changeButtonStatusIfNeeded();
+    changeButtonStatus();
   });
 
 
 }
 
-function changeButtonStatusIfNeeded() {
+function changeButtonStatus() {
   recordButton.disabled = isRecording ? true : false;
   stopButton.disabled = isRecording ? false: true;
   downloadButton.disabled = isRecording ? true: false;
@@ -262,7 +260,7 @@ function setupRecordingGainVis() {
   const amplitude = height * 2;
 
   function draw(currentSampleGain) {
-    const centerY = height / 2 - currentSampleGain * amplitude;
+    const currentY = height / 2 - currentSampleGain * amplitude;
 
     // Clear current Y-axis.
     canvasContext.clearRect(currentX, 0, 1, height);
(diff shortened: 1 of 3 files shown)


src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -15,20 +15,26 @@ const WAVEFROM_SCALE_FACTOR = 5
...
-document.addEventListener('click', (element) => {
-  init();
-  document.querySelector('#click-to-start').remove();
-}, {once: true});
+if (!initCount){
+  recordButton.disabled = false;
+  recordButton.addEventListener('click', (element) => {
+    init();
+    isRecording = true;
+    initCount++;
```
hoch, line 33: Why do we need counting? Can't this be `_isInitialized` boolean flag?

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -121,7 +124,7 @@ function setupScriptProcessor(recordingProperties, passSampleToVisualizers) {
       
       if (recordingLength > recordingProperties.maxFrameCount) {
         isRecording = !isRecording;
-
+        window.alert("The recording length reach the max limit!");
```
hoch, line 127: I advise against using `alert`. Probably stopping and displaying some warning texts in `div` would be better.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -147,17 +149,25 @@ function setupScriptProcessor(recordingProperties, passSampleToVisualizers) {
...
+    isRecording = false;
+    const finalRecordBuffer = createFinalRecordBuffer(recordingProperties);
+    prepareClip(finalRecordBuffer);
+    changeButtonDisabled();
   });
+
+
+}
+
+function changeButtonDisabled() {
```
hoch, line 167: The naming scheme in Chromium would be:
- `changeButtonStatusIfNeeded()`
- `maybeChangeButtonStatus()`

This suggests: this function may or may not change the status of buttons based on some factors.

src/audio-worklet/migration/worklet-recorder/app.js
```diff
@@ -11,13 +11,25 @@ const context = new AudioContext();
 // Make the visulization more clear to the users
 const WAVEFROM_SCALE_FACTOR = 5
 let isRecording = false;
-let visualizationEnabled = true;
+let initCount = 0;
```
hoch, line 14: Ditto.

src/audio-worklet/migration/worklet-recorder/app.js
```diff
@@ -11,13 +11,25 @@ const context = new AudioContext();
...
-  document.querySelector('#click-to-start').remove();
-}, {once: true});
+if (!initCount){
+  recordButton.disabled = false;
+  recordButton.addEventListener('click', (element) => {
+    init();
+    isRecording = true;
+    changeButtonDisabled();
+    initCount++;
+    recordText.textContent = "Continue";
```
hoch, line 30: nit: Use single quotes for the strings.

hoch (commented):
Overall it looks good. Can you add "before/after" screenshots?

src/audio-worklet/migration/spn-recorder/app.js
TenghuiZhang: I make the count here instead of the boolean is that I want to try my best keep the SPN recorder and worklet recorder using the same parameter so the developer can have a better and clear compare the difference between these two method.

src/audio-worklet/migration/worklet-recorder/app.js
TenghuiZhang: The reason why I use count instead of boolean is that, we need to control the first time init is sending the "message: 'UPDATE_RECORDING_STATE',  to processor. Since the workflow is: "Click start recording" -> isInitied = true -> Then we don't have another control variable or using other control variable to control  the message posting to audioworklet processor. Therefore, I use initCount as a number to control these two steps. Sorry for the confusion. Maybe I should use a better naming?

TenghuiZhang:
This is the old sound wave design: 
Old SPN design

This is the new sound wave design when it is recording:
SPN finish

This is the new alert message when the recording reach the max time limit:
SPN recording

src/audio-worklet/migration/worklet-recorder/app.js
hoch: Don't be sorry! You're doing great here.

If we need to handle more than a binary state in the code, it's better to use the enumeration to make it more explicit. How many states do we need in the SPN example?

src/audio-worklet/migration/spn-recorder/app.js
hoch: See my response below.

src/audio-worklet/migration/worklet-recorder/app.js
TenghuiZhang: Thanks for the advices! I split SPN recording state into three:
const recordingStates = {
  notInitialized: 'notInitialized',
  startIniAndRecording: 'startIniAndRecording',
  finishRecording: 'finishRecording'
};

And Worklet recording state into four since Worklet has one more state of sending message to recording-processor:
const recordingStates = {
  notInitialized: 'notInitialized',
  startInitialized: 'startInitialized',
  startRecording: 'startRecording',
  finishRecording: 'finishRecording'
};

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -6,6 +6,13 @@
 
 import createLinkFromAudioBuffer from './exporter.mjs';
 
+// This enum states the current recording state
```
hoch, line 9: Nice! How about this?

```js
const RecorderState = {
  UNINITIALIZED: 0,
  RECORDING: 1,
  FINISHED: 2,
};
```

Then you can use it like:

```js
if (recorderState === RecorderState.UNINITIALIZED) {
  ...
}
```

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -15,20 +22,28 @@ const WAVEFROM_SCALE_FACTOR = 5
...
 let downloadLink = document.querySelector('#download-link');
 
+
+
 // Wait for user interaction to initialize audio, as per specification.
-document.addEventListener('click', (element) => {
-  init();
-  document.querySelector('#click-to-start').remove();
-}, {once: true});
+if (recordingState === recordingStates.notInitialized){
```
hoch, line 37: nit: a space between `){`.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -307,32 +257,41 @@ function setupRecordingGainVis() {
...
+  let previousY = height / 2;
+  // Adjust the amplitude value to increase or decrease the size of the waveform
+  const amplitude = height * 2;
 
   function draw(currentSampleGain) {
-    const centerY = ((1 - currentSampleGain) * height) / 2;
-    const gainHeight = currentSampleGain * height;
+    const centerY = height / 2 - currentSampleGain * amplitude;
 
     // Clear current Y-axis.
```
hoch, line 267: I don't find these comments (around `canvas`) are useful or meaningful. Let's get rid of them.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -307,32 +257,41 @@ function setupRecordingGainVis() {
...
     canvasContext.clearRect(currentX, 0, 1, height);
 
     // Draw recording bar 1 ahead.
```
hoch, line 270: Ditto.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -307,32 +257,41 @@ function setupRecordingGainVis() {
...
     canvasContext.fillStyle = 'red';
-    canvasContext.fillRect(currentX+1, 0, 1, height);
+    canvasContext.fillRect(currentX + 1, 0, 1, height);
+
+    // Draw line plot.
```
hoch, line 274: Ditto.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -307,32 +257,41 @@ function setupRecordingGainVis() {
...
+    canvasContext.beginPath();
+    canvasContext.moveTo(currentX, previousY);
+    canvasContext.lineTo(currentX + 1, centerY);
+    canvasContext.strokeStyle = 'black';
+    // Decrease the line width for better visibility
```
hoch, line 279: Ditto.

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -307,32 +257,41 @@ function setupRecordingGainVis() {
...
+    canvasContext.lineWidth = 0.8;
+    canvasContext.stroke();
 
-    // Draw current gain.
```
hoch, line 322: Ditto.

hoch (commented):
A couple of comments, but we're on the right direction.

hoch:
Yes. The new screenshot looks a lot better than the previous version. Can we increase the width of the volume level meter? (5~7 pixels?)

hoch (approved):
LGTM

Loved some details in the change! :)

Issue #332, opened 2023-05-27
Implement a volume meter on the right side of waveform display.


TenghuiZhang:
We need to implement a volume meter on the right side of waveform

PR #335, opened 2023-06-02
CanvasMeter

TenghuiZhang:
This PR is fixing #332

### DIFF
--- a/src/audio-worklet/migration/spn-recorder/app.js
+++ b/src/audio-worklet/migration/spn-recorder/app.js
@@ -13,7 +13,7 @@ const BUFFER_SIZE = 256;
 // Make the visulization more clear to the users
 const WAVEFROM_SCALE_FACTOR = 5
 // Make the visulization of vu meter more clear to the users
-const VU_METER_SCALE_FACTOR = 4000
+const VU_METER_SCALE_FACTOR = 1000
 let recordingLength = 0;
 let recordBuffer = [[], []];
 let isRecording = false;
--- a/src/audio-worklet/migration/spn-recorder/app.js
+++ b/src/audio-worklet/migration/spn-recorder/app.js
@@ -259,7 +259,7 @@ function setupRecordingGainVis() {
 
   let currentX = 0;
   let previousY = height / 2;
-  // Adjust the amplitude value to increase or decrease the size of the waveform
+  // Adjust the amplitude value to increase the size of the waveform
   const amplitude = height * 2;
 
   function draw(currentSampleGain) {
--- a/src/audio-worklet/migration/spn-recorder/app.js
+++ b/src/audio-worklet/migration/spn-recorder/app.js
@@ -253,7 +253,7 @@ function setupRecordingGainVis() {
   let currentX = 0;
   let previousY = height / 2;
   // Adjust the amplitude value to increase or decrease the size of the waveform
-  const amplitude = height * 1.5;
+  const amplitude = height;
 
   function draw(currentSampleGain) {
     const centerY = height / 2 - currentSampleGain * amplitude;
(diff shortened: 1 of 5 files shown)


TenghuiZhang:
This is the design of canvas vu-meter
canvas-meter

TenghuiZhang:
Padding fixed screenshot

VM meter

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -252,11 +258,17 @@ function setupRecordingGainVis() {
...
   function draw(currentSampleGain) {
-    const currentY = height / 2 - currentSampleGain * amplitude;
+    // This formula is design based on this logic:
+    // Middle line of canvas: height / 2
+    // Current sound wave gain value range is -1 to 1
+    // We want use current gain value divide by gain value range and
+    // time half of canvas height, therefore, we can get the
+    // accurate wave size.
+    // At the end, use scale_factor to make is clearer for users
+    const currentY = height / 2 - height / 2 * (currentSampleGain
```
hoch, line 270: nit: Perhaps we can rewrite this -

```js
const currentY =
    height / 2 * (1 - currentSampleGain / GAIN_NODE_MAX_VALUE) 
    * SCALE_FACTOR;
```

src/audio-worklet/migration/spn-recorder/app.js
```diff
@@ -18,8 +18,11 @@ const context = new AudioContext();
 
 // Arbitrary buffer size, not specific for a reason
 const BUFFER_SIZE = 256;
-// Make the visulization more clear to the users
-const WAVEFROM_SCALE_FACTOR = 5;
+// Make the visulization clearer to the users
+const SCALE_FACTOR = 10;
+// Make the visulization of vu meter more clear to the users
+const GAIN_NODE_MAX_VALUE = 1;
```
hoch, line 24: `MAX_GAIN = 1` should be straightforward enough.

hoch (approved):
LGTM with nits

src/audio-worklet/migration/spn-recorder/app.js
TenghuiZhang: I think you means: const currentY =
    height / 2 - (1 - currentSampleGain / GAIN_NODE_MAX_VALUE) 
    * SCALE_FACTOR;
    
 Since we are drawing the the waveform from the center of the canvas. 

But this visualization will not be as good as the previous formula. Therefore, I will keep the original one.

hoch: I simply factored the math formula you (because `h/2` is a common factor) had and rewrapped it. I don't think I changed the formula itself. Actually it can even be simplified further:

```js
const currentY = height / 2 * (1 - currentSampleGain * SCALE_FACTOR);
```

Because `MAX_GAIN` is always 1. Right?

TenghuiZhang: Gotcha. Thanks for the explanation! But currentSampleGain is a value between -1 to 1. So I assume it will be: 

```
const currentY = height / 2* (1 - (currentSampleGain
        / MAX_GAIN) * SCALE_FACTOR);
```

hoch: Does `MAX_GAIN`change in the code somewhere? I don't think it's possible because it's `const`. Am I missing something?

If you're thinking about keeping `MAX_GAIN` for readability I am okay with that, but still you're practically dividing something with 1. I was pointing out that's redundant.

TenghuiZhang: Okay! I will remove the Max_gain

```
 const currentY = height / 2 * (1 - currentSampleGain * SCALE_FACTOR);
```

PR #346, opened 2023-10-12
fixed ring buffer indexes updates in design patterns

grazder:
Fixed RingBuffer read / write indexes update.
Because of that error i faced lags during AudioWorklet process because of that.
Faced when RingBuffer length 48000 and kernelBufferSize - 512


mjwilson-google:
Thank you!  The fix looks correct to me, but we can't accept your contribution unless we have a CLA on file.  Please comment here when the CLA is ready, or if you need help sorting this out.

grazder:
I think it's ready

mjwilson-google (approved):

Issue #347, opened 2023-10-13
Improve RingBuffer


hoch:
A comment from @mjwilson-google: 

> The way we're reading or writing the samples is also kinda strange; we could probably just be updating the index directly instead of making a new temporary every iteration. But that at least looks like it's correct, just inefficient. If we updated the index directly each iteration we wouldn't need to do the correction after the loop at all.

I say we should prioritize improving our AudioWorklet examples because it is often used/referred by developers.

mjwilson-google:
This was referencing https://github.com/GoogleChromeLabs/web-audio-samples/blob/main/src/audio-worklet/design-pattern/lib/wasm-audio-helper.js but it seems like we have some other ring buffer implementations as well that might be worth looking at too.

Daksh-10:
@hoch @mjwilson-google I worked on this issue and created a pull request #352

hoch:
This is a tracking issue for improving the RingBuffer. The PR above can be considered one task that we completed.

hoch:
@professorabhay is working on this:
https://github.com/GoogleChromeLabs/web-audio-samples/issues?q=is%3Aopen+author%3Aprofessorabhay

PR #352, opened 2024-02-25
Fix: #347 : Improve RingBuffer

Daksh-10:
In https://github.com/GoogleChromeLabs/web-audio-samples/issues/347 ,
I updated the index directly instead of making a new temporary every iteration, which removes the inefficiency.

### DIFF
--- a/src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
+++ b/src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
@@ -216,13 +216,11 @@ class RingBuffer {
     // Transfer data from the |arraySequence| storage to the internal buffer.
     const sourceLength = arraySequence[0].length;
     for (let i = 0; i < sourceLength; ++i) {
-      const writeIndex = (this._writeIndex + i) % this._length;
+      this._writeIndex = (this._writeIndex + i) % this._length;
       for (let channel = 0; channel < this._channelCount; ++channel) {
-        this._channelData[channel][writeIndex] = arraySequence[channel][i];
+        this._channelData[channel][this._writeIndex] = arraySequence[channel][i];
       }
     }
-
-    this._writeIndex = (this._writeIndex + sourceLength) % this._length;
 
     // For excessive frames, the buffer will be overwritten.
     this._framesAvailable += sourceLength;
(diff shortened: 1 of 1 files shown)


src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -216,14 +216,12 @@ class RingBuffer {
     // Transfer data from the |arraySequence| storage to the internal buffer.
     const sourceLength = arraySequence[0].length;
     for (let i = 0; i < sourceLength; ++i) {
-      const writeIndex = (this._writeIndex + i) % this._length;
+      this._writeIndex = (this._writeIndex + i) % this._length;
```
mjwilson-google, line 219: We should just be incrementing `this._writeIndex` by `1` each iteration now.

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -249,14 +247,12 @@ class RingBuffer {
 
     // Transfer data from the internal buffer to the |arraySequence| storage.
     for (let i = 0; i < destinationLength; ++i) {
-      const readIndex = (this._readIndex + i) % this._length;
+      this._readIndex = (this._readIndex + i) % this._length;
```
mjwilson-google, line 250: Same as above, we should just be incrementing by `1` instead of `i` now that we're storing back to the class variable.

mjwilson-google (changes requested):
Thank you for working on this!  Please address the comments in the code.  It would also be good to run src/audio-worklet/design-pattern/wasm-ring-buffer/main.js before and after the change and verify that this change doesn't affect the behavior.

Daksh-10:
@mjwilson-google,
 I corrected the changes.

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -216,14 +216,13 @@ class RingBuffer {
     // Transfer data from the |arraySequence| storage to the internal buffer.
     const sourceLength = arraySequence[0].length;
     for (let i = 0; i < sourceLength; ++i) {
-      const writeIndex = (this._writeIndex + i) % this._length;
+      //updating index directly instead of making new temporary every iteration
```
mjwilson-google, line 219: I don't think this comment is necessary.

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -249,14 +248,13 @@ class RingBuffer {
 
     // Transfer data from the internal buffer to the |arraySequence| storage.
     for (let i = 0; i < destinationLength; ++i) {
-      const readIndex = (this._readIndex + i) % this._length;
+      //updating index directly instead of making new temporary every iteration
```
mjwilson-google, line 251: Same here, since the old line is being removed the context for the comment is lost.  We can just make the change without adding this comment.

mjwilson-google (changes requested):
Thanks for updating.  I missed this in the initial review, but we need to increment the indexes after using them now, not before.  Once we do this I think everything should be good!

Daksh-10:
I think we took time to correct the issue, but I guess it is working correctly now.

mjwilson-google (approved):

PR #355, opened 2024-03-29
optimize buffer access in the RingBuffer class

Shivam7-1:
Optimize Buffer Access in RingBuffer Class
Issue: https://github.com/GoogleChromeLabs/web-audio-samples/issues/347
Description:

This PR aims to optimize buffer access in the RingBuffer class by utilizing typed array views for more efficient data access. The changes improve performance by avoiding nested array accesses and directly accessing the underlying Float32Array.

Changes:

- Refactored buffer access methods (`push` and `pull`) in the RingBuffer class to utilize typed array views for accessing channel data.
- Removed nested array accesses and directly accessed the `_data` property of the RingBuffer class for improved efficiency.
- Simplified index calculations for accessing channel data.

### DIFF
--- a/src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
+++ b/src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
@@ -189,11 +189,13 @@ class RingBuffer {
 
     this._channelCount = channelCount;
     this._length = length;
-    this._channelData = [];
-    for (let i = 0; i < this._channelCount; ++i) {
-      this._channelData[i] = new Float32Array(length);
-    }
+    this._data = new Float32Array(length * channelCount);
   }
+
+  get framesAvailable() {
+    return this._framesAvailable;
+  }
+
 
   /**
    * Getter for Available frames in buffer.
@@ -215,20 +217,23 @@ class RingBuffer {
 
     // Transfer data from the |arraySequence| storage to the internal buffer.
     const sourceLength = arraySequence[0].length;
+    const dataArray = this._data;
+    const channelCount = this._channelCount;
+    const length = this._length;
+
     for (let i = 0; i < sourceLength; ++i) {
-      for (let channel = 0; channel < this._channelCount; ++channel) {
-        this._channelData[channel][this._writeIndex] = arraySequence[channel][i];
+      for (let channel = 0; channel < channelCount; ++channel) {
+        const index = (this._writeIndex + i) % length;
+        dataArray[index * channelCount + channel] = arraySequence[channel][i];
       }
-      this._writeIndex = (this._writeIndex + 1) % this._length;
     }
 
-    // For excessive frames, the buffer will be overwritten.
+    this._writeIndex = (this._writeIndex + sourceLength) % length;
     this._framesAvailable += sourceLength;
-    if (this._framesAvailable > this._length) {
-      this._framesAvailable = this._length;
+    if (this._framesAvailable > length) {
+      this._framesAvailable = length;
     }
   }
-
   /**
    * Pull data out of buffer and fill a given sequence of Float32Arrays.
    *
@@ -244,15 +249,18 @@ class RingBuffer {
     }
 
     const destinationLength = arraySequence[0].length;
+    const dataArray = this._data;
+    const channelCount = this._channelCount;
+    const length = this._length;
 
-    // Transfer data from the internal buffer to the |arraySequence| storage.
     for (let i = 0; i < destinationLength; ++i) {
-      for (let channel = 0; channel < this._channelCount; ++channel) {
-        arraySequence[channel][i] = this._channelData[channel][this._readIndex];
+      for (let channel = 0; channel < channelCount; ++channel) {
+        const index = (this._readIndex + i) % length;
+        arraySequence[channel][i] = dataArray[index * channelCount + channel];
       }
-      this._readIndex = (this._readIndex + 1) % this._length;
     }
 
+    this._readIndex = (this._readIndex + destinationLength) % length;
     this._framesAvailable -= destinationLength;
     if (this._framesAvailable < 0) {
       this._framesAvailable = 0;


Shivam7-1:
Hi @mjwilson-google 

I've made some significant changes in PR #347 , where I optimized buffer access in the RingBuffer class. I believe these changes could improve the performance of our project.

Could you please take a moment to review my PR? Your feedback would be greatly appreciated.

Additionally, I'm planning to apply for GSOC'24 and I'm particularly interested in this project. I believe that your insights would not only help improve my PR but also my application for GSOC'24.

Thank you for your time and consideration.

Best,
Shivam

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -189,12 +189,14 @@ class RingBuffer {
...
-    this._channelData = [];
-    for (let i = 0; i < this._channelCount; ++i) {
-      this._channelData[i] = new Float32Array(length);
-    }
+    this._data = new Float32Array(length * channelCount);
+  }
+
+  get framesAvailable() {
+    return this._framesAvailable;
   }
```
mjwilson-google, line 197: This looks like a duplicate of the function below; was this intended?

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -215,20 +217,23 @@ class RingBuffer {
 
     // Transfer data from the |arraySequence| storage to the internal buffer.
     const sourceLength = arraySequence[0].length;
+    const dataArray = this._data;
+    const channelCount = this._channelCount;
+    const length = this._length;
```
mjwilson-google, line 222: Are these const aliases for performance purposes?  If so, did you measure the performance difference versus accessing the object members directly?

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -244,15 +249,18 @@ class RingBuffer {
     }
 
     const destinationLength = arraySequence[0].length;
+    const dataArray = this._data;
+    const channelCount = this._channelCount;
+    const length = this._length;
 
-    // Transfer data from the internal buffer to the |arraySequence| storage.
```
mjwilson-google, line 248: Why remove this comment?

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -215,20 +217,23 @@ class RingBuffer {
...
-      for (let channel = 0; channel < this._channelCount; ++channel) {
-        this._channelData[channel][this._writeIndex] = arraySequence[channel][i];
+      for (let channel = 0; channel < channelCount; ++channel) {
+        const index = (this._writeIndex + i) % length;
+        dataArray[index * channelCount + channel] = arraySequence[channel][i];
       }
-      this._writeIndex = (this._writeIndex + 1) % this._length;
     }
 
-    // For excessive frames, the buffer will be overwritten.
```
mjwilson-google, line 225: I think we need to keep some form of this comment, and actually looking at this a bit more closely we probably need to update the read index if we overflow the buffer.

Some ringbuffers will stop when they get full; this one just keeps writing.  Changing that is probably outside the scope of this PR.  But we should probably at least keep the comment.

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
```diff
@@ -215,20 +217,23 @@ class RingBuffer {
...
     const sourceLength = arraySequence[0].length;
+    const dataArray = this._data;
+    const channelCount = this._channelCount;
+    const length = this._length;
+
     for (let i = 0; i < sourceLength; ++i) {
-      for (let channel = 0; channel < this._channelCount; ++channel) {
-        this._channelData[channel][this._writeIndex] = arraySequence[channel][i];
+      for (let channel = 0; channel < channelCount; ++channel) {
+        const index = (this._writeIndex + i) % length;
```
mjwilson-google, line 226: Did you measure that this is more efficient than storing back to the object member?  There was a recent change that did the opposite of what is being proposed here so it would be good to quantify which version is faster, and maybe add a comment with the results as well.

mjwilson-google (commented):
Thank you for the pull request!

I have added some comments.  I think making the array contiguous should improve performance but it would be best if we had some profiling or other measurement to make sure everything is having the effect we desire.

There are some strange things about these ringbuffer implementations so I appreciate your interest and wish you luck on your application!

src/audio-worklet/design-pattern/lib/wasm-audio-helper.js
Shivam7-1: You're right, keeping some form of the comment is important to acknowledge the discussion about the efficiency of the proposed optimization. Additionally, updating the read index if the buffer overflows is a valid consideration, as it ensures proper behavior when the buffer is full. While changing the behavior of the ring buffer to stop when it gets full might be outside the scope of this PR, updating the comment and considering the overflow scenario are reasonable adjustments. I'll make sure to address both points in the code and update the comment accordingly. Thank you for your input!

Shivam7-1: Hi
I had some Approach to  updated code:

-  After pushing data into the buffer, the write index and frames available are updated as before.
- Additionally, if the frames available exceed the buffer length, the read index is updated to the current write index, effectively wrapping around the buffer.
- The pull method remains unchanged, only updating the read index and frames available.

mjwilson-google: This sounds fine to me.  I think it may be more common in audio to stop writing when full, probably because that would produce one large discontinuity instead of potentially many small discontinuities, but as long as we are clear about what happens when more data is written than there is space in the buffer either way should be OK.

The pull method has a similar issue of the read index overtaking the write index if we read more than _framesAvailable.

hoch:
I am closing this because a similar task is completed by https://github.com/GoogleChromeLabs/web-audio-samples/pull/369. Thank you for your interest and contribution!

PR #360, opened 2024-05-14
Add 4 descriptions to landing page demos

terryzfeng:
Noticed that Wavetable Synthesizer, AnalyserNode with visualization, Panning and Reverberation, and WebGL Pool with 3d audio didn't have descriptions so added them. Perhaps might be more than a non-functional change.


mjwilson-google (approved):
LGTM

PR #366, opened 2024-05-17
Combined the classes from wasm-audio-helper.js in a new file

professorabhay:
### Combined Classes - 

- Created a new file `free-queue.js` in `lib`.
- Combined `HeapAudioBuffer` and `RingBuffer` from `wasm-audio-helper.js`  and  `FreeQueue` class from `free-queue.js` into a single class `FreeQueue` in the above file.
- Removed the existing `free-queue.js` file from `free-queue\src\`.
- Moved the `free-queue` folder to `design-patterns`, as it consists of the example files of FreeQueue.
- Updated all the files that use the methods of these classes to the latest.


src/audio-worklet/design-pattern/free-queue/examples/simple-passthrough/basic-processor.js
```diff
@@ -1,4 +1,4 @@
-import FreeQueue from "../../src/free-queue.js";
+import {RingBufferSharedArrayBuffer} from "../../../design-pattern/lib/ring-buffer.js";
```
hoch, line 1: I am sorry, but can we keep the class name `FreeQueue`?

src/audio-worklet/design-pattern/lib/ring-buffer.js
```diff
@@ -1,3 +1,250 @@
...
+    this._maxChannelCount = maxChannelCount ?
+        Math.min(maxChannelCount, MAX_CHANNEL_COUNT) : channelCount;
+    this._channelCount = channelCount;
+    this._allocateHeap();
+    this._isInitialized = true;
+
+    this._readIndex = 0;
+    this._writeIndex = 0;
+    this._framesAvailable = 0;
+
```
hoch, line 60: nit: remove a blank line

src/audio-worklet/design-pattern/lib/ring-buffer.js
```diff
@@ -1,3 +1,250 @@
...
+   * Frees the allocated memory space in the WASM heap.
+   */
+  free() {
+    this._isInitialized = false;
+    this._module._free(this._dataPtr);
+    this._module._free(this._pointerArrayPtr);
+    this._channelData = null;
+  }
+
+/**
```
hoch, line 169: This comment's location seems awkward. Shouldn't this be at the top?

src/audio-worklet/design-pattern/lib/ring-buffer.js
```diff
@@ -1,3 +1,250 @@
...
+  get framesAvailable() {
+    return this._framesAvailable;
+  }
+
+  /**
+   * Push a sequence of Float32Arrays to buffer.
+   *
+   * @param  {array} arraySequence A sequence of Float32Arrays.
+   */
+  pushRing(arraySequence) {
```
hoch, line 197: Not sure about the name of this method. I think we prefer "verb + noun", but here the noun part is unclear. What's "ring"? Being explicit always helps - `pushArraySequence`.

Also it looks like you're pushing `sample-frames` - a packet of multi-channel audio stream

src/audio-worklet/design-pattern/lib/ring-buffer.js
```diff
@@ -1,3 +1,250 @@
...
+      this._framesAvailable = this._length;
+    }
+  }
+
+  /**
+   * Pull data out of buffer and fill a given sequence of Float32Arrays.
+   *
+   * @param  {array} arraySequence An array of Float32Arrays.
+   */
+  pullRing(arraySequence) {
```
hoch, line 222: Ditto.

hoch (changes requested):
Names of variable/class are a bit confusing. We might want to build a consensus before moving forward.

src/audio-worklet/design-pattern/free-queue/examples/simple-passthrough/main.js
```diff
@@ -1,12 +1,12 @@
-import {RingBufferSharedArrayBuffer} from "../../../design-pattern/lib/ring-buffer.js";
+import {FreeQueue} from "../../../lib/free-queue.js";
 import { QUEUE_SIZE } from './constants.js';
```
hoch, line 2: `{QUEUE_SIZE}`

Could you fix this while we're at it?

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -25,6 +25,17 @@ const MAX_CHANNEL_COUNT = 32;
 // WebAudio's render quantum size.
 const RENDER_QUANTUM_FRAMES = 128;
 
+/**
+ * A shared storage for FreeQueue operation backed by SharedArrayBuffer.
```
hoch, line 29: I don't see the matching implementation of this comment. Where is it?

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -33,33 +44,92 @@ const RENDER_QUANTUM_FRAMES = 128;
  * @class
  * @dependency Module A WASM module generated by the emscripten glue code.
  */
-class RingBufferWasmHeap{
+class FreeQueue{
```
hoch, line 47: We need a space between the class name and `{`.

I recommend running the included linter against this file. See this command line:
https://github.com/GoogleChromeLabs/web-audio-samples/blob/main/package.json#L14

You can use `npx eslint` command line to target a specific file.

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -33,33 +44,92 @@ const RENDER_QUANTUM_FRAMES = 128;
...
+
+    /**
```
hoch, line 49: incorrect indentation (from here and below)

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -33,33 +44,92 @@ const RENDER_QUANTUM_FRAMES = 128;
...
+   */
+    States = {
+      /** @type {number} A shared index for reading from the queue. (consumer) */
+      READ: 0,
+      /** @type {number} A shared index for writing into the queue. (producer) */
+      WRITE: 1,  
+    }
+
+    /**
+   * Constructor
```
hoch, line 61: First, the comment here doesn't say anything about different arguments and how the constructor can behave differently. Please fix the comment and explain the new mechanism.

Also please consider using `option bag` approach for this. For example:

`new FreeQueue({bufferSize: 512, channelCount: 2})`

or 

`new FreeQueue({wasmModule: m, bufferSize: 512, channelCount: 2, maxChannelCount: 8})`

This way you don't need type check for every argument; you can simply check if the given option bag has a certain field to figure out the behavior. Actually in this case, all you need to check is `wasmModule`.

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -33,33 +44,92 @@ const RENDER_QUANTUM_FRAMES = 128;
...
+        throw new Error("Invalid constructor arguments");
+      }
+    }
+
+     /**
+   * Initialize as a SharedRingBuffer
+   * @param {number} size
+   * @param {number} channelCountRing
+   */
+  _initSharedRingBuffer(size, channelCountRing=1) {
```
hoch, line 85: Let's spell out when it's possible - `_initializeRingBuffer`

Also what's `channelCountRing`?

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -33,33 +44,92 @@ const RENDER_QUANTUM_FRAMES = 128;
...
+    this.states = new Uint32Array(
```
hoch, line 86: You can write

```js
this.states = new Uint32Array(new SharedArrayBuffer(
    Object.keys(this.States).length * Uint32Array.BYTES_PER_ELEMENT));
```

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -33,33 +44,92 @@ const RENDER_QUANTUM_FRAMES = 128;
...
+      )
+    );
+    /**
+     * Use one extra bin to distinguish between the read and write indices 
+     * when full. See Tim Blechmann's |boost::lockfree::spsc_queue|
+     * implementation.
+     */
+    this.bufferLength = size + 1;
+    this.channelCountRing = channelCountRing;
+    this.channelDataRing = [];
```
hoch, line 98: Suffixing the variables with `~Ring` doesn't seem like a good approach.

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -33,33 +44,92 @@ const RENDER_QUANTUM_FRAMES = 128;
...
+    for (let i = 0; i < channelCountRing; i++) {
+      this.channelDataRing.push(
```
hoch, line 100: Same here: please take a look at my example above to make the indentation correct.

For continuing lines, you need 4 spaces:
https://google.github.io/styleguide/jsguide.html#formatting-indent

src/audio-worklet/design-pattern/lib/free-queue.js
```diff
@@ -33,33 +44,92 @@ const RENDER_QUANTUM_FRAMES = 128;
...
+            this.bufferLength * Float32Array.BYTES_PER_ELEMENT
+          )
+        )
+      );
+    }
+  }
 
   /**
-   * @constructor
+   * Initialize as a WASM HEAP wrapper
```
hoch, line 111: Was there any original comment for this method? If not, we would need a new one. The constructor's comment is important. We need to say what this class does and probably with some usage examples.

src/audio-worklet/design-pattern/wasm-supersaw/coi-serviceworker.js
```diff
@@ -0,0 +1,115 @@
+/*! coi-serviceworker v0.1.6 - Guido Zuidhof, licensed under MIT */
```
hoch, line 1: I think we need this file under `src/third_party/` and all the examples should refer that single file.

src/audio-worklet/design-pattern/lib/free-queue.js
hoch, line 1: We need to print an error if the SharedArrayBuffer is not supported. (e.g. COOP/COEP)

src/audio-worklet/design-pattern/wasm/coi-serviceworker.js
hoch, line 1: Ditto.

hoch (changes requested):
It's a good start!

I see some concerns around unifying two models into one. Some thoughts:
1. The SharedArrayBuffer version will not run on the constrained domain. (e.g. GitHub)
2. It might be better to keep 2 different version of the `FreeQueue` class
  - FreeQueue (single thread version)
  - FreeQueueSAB (multi-thread support version)

My suggestion is not to touch the files under `audio-worklet/design-pattern`, and iterate a new class under the `src/lib/free-queue` directory.

I also do not see any worker-driven demo (`shared-buffer`) in this PR. Is there any reason for that? If you couldn't get it working, it means that we need a different approach like I mentioned above.

src/audio-worklet/design-pattern/wasm-supersaw/coi-serviceworker.js
professorabhay: Ah, better approach! Thanks for pointing out. 
I will make the changes.

src/audio-worklet/design-pattern/lib/free-queue.js
professorabhay: The code related to the old `free-queue` class starts from line `317`.

PR #369, opened 2024-06-16
Consolidating the existing RingBuffer Implementation.

professorabhay:
- Created a new  `lib/free-queue/` directory under `src/`
- Added `free-queue.js` and `free-queue-sab.js` files 
- Added the code
- updated the imports/usage with new classes


mjwilson-google (changes requested):
Thanks for the change!

When I tested this, I needed to add the following to the "  // Passthrough files via these glob patterns." list in .eleventy.js:
    'src/lib/**/*.js',

Once I did that, the four examples worked as expected.  Please include this change and I will approve the PR.

mjwilson-google (approved):

PR #373, opened 2024-07-10
Downgrade eslint to be compatible with eslintrc

Kizjkre:
Downgrade eslint from 9.5.0 to 8.57.0 to be compatible with eslintrc.


hoch:
I found [this](https://eslint.org/docs/latest/use/configure/#configuration-file-formats), but couldn't find any reference that points the version 8.57.0. How did you find it out? :)

hoch (approved):
LGTM

PR #374, opened 2024-07-10
FreeQueueSAB: automated test coverage

professorabhay:
- Installed needed package - Mocha and Chai ✅
- Folder Structure Setup ✅
- Test cases for FreeQueueSAB Class  ✅

Fixes https://github.com/GoogleChromeLabs/web-audio-samples/issues/347


hoch:
The PR at the moment only adds two packages. Perhaps it's not ready for the review yet? :)

professorabhay:
> The PR at the moment only adds two packages. Perhaps it's not ready for the review yet? :)

Hey,  @hoch! Hope you are doing well. 
Yes, it's not ready for the review yet. 
I have a last exam today. After that I'll be free from tomorrow and push the code soon. I have created the Test Cases for the FreeQueueSAB class.

src/lib/free-queue/test.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>FreeQueueSAB Tests</title>
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/mocha/8.4.0/mocha.min.css">
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/mocha/8.4.0/mocha.min.js"></script>
```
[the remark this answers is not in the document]
hoch: Hmm. Can we source this from another CDN? Not sure why GitHub thinks Cloudflare is an untrusted source.

src/lib/free-queue/test.html
```diff
@@ -0,0 +1,28 @@
...
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/chai/4.3.4/chai.min.js"></script>
```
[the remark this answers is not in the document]
hoch: Ditto.

src/lib/free-queue/test/test-free-queue-sab.js
hoch, line 1: It looks like this test is targeting the buffer index operation. I defer my review to @mjwilson-google.

hoch (commented):
Looks generally good, but I assume this is the first PR of many. Right?

src/lib/free-queue/test.html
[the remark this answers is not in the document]
professorabhay: Hmm, I am also confused why this is happening. I'll figure out something.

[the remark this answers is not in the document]
mjwilson-google: Seems like we can add a resource integrity check for the external files:
https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity

professorabhay:
> Looks generally good, but I assume this is the first PR of many. Right?

As per the discussion in the last meeting, we will cover all the test cases related to the FreeQueueSAB class in this PR and create a new PR for all test cases related to the FreeQueue class. 

I'll be committing more test cases related to performance testing soon. Please provide me a review and any case that is not covered according to you so that we can move towards finalizing this PR.

src/lib/free-queue/test.html
```diff
@@ -0,0 +1,28 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>FreeQueueSAB Tests</title>
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/mocha/10.6.0/mocha.min.css" integrity="sha384-BBZj3VfQ5pAok2bd0uXf+8cRAJ3D/JZWrxBlnWx6BcsBO96ZzZR40fpC20MtWbM8" crossorigin="anonymous">
```
mjwilson-google, line 7: When I tried to test this locally, I got an error that the hash didn't match.

I think we can just copy the SHA-512 hash directly from cdnjs:
https://cdnjs.com/libraries/mocha
(click the shield icon next to the filename)

src/lib/free-queue/test.html
```diff
@@ -0,0 +1,28 @@
...
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/mocha/10.6.0/mocha.min.js" integrity="sha384-A5EqeN0zznOH0u6V4OduoPlHhkhk+Km4eBBNBq0UAE1Z51cFJfX/f9biOo2SmCCS" crossorigin="anonymous"></script>
```
mjwilson-google, line 8: Same issue with the hash as above.

src/lib/free-queue/test.html
```diff
@@ -0,0 +1,28 @@
...
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/chai/5.1.1/chai.min.js" integrity="sha384-Zg6S9cOeB+JIKb63Z7Qi7N1vGb1nctA+lPjPBmAxFQf3mxxu1BO+e+b62MkzJeEt" crossorigin="anonymous"></script>
```
mjwilson-google, line 9: chai.min.js doesn't seem to be available on cdnjs for me.  I just see the full chai.js:
https://cdnjs.com/libraries/chai

Trying to use chai.js I get the error `SyntaxError: Unexpected token 'export'` in chai.js when loading the page.

mjwilson-google (changes requested):
I'm getting some errors when trying to run the tests.  Please take a look.

Is it possible to include the necessary mocha and chai support files locally, or implement the parts we need separately?

Also we may need to include the following line in .eleventy.js to include the test html file in the project:
    'src/lib/**/*.html',

src/lib/free-queue/test.html
professorabhay: Yes, I resolved it. After the meeting, I tried to re-run the test cases and that error popped up.

src/lib/free-queue/test.html
professorabhay: Yes, I used an external hash generator site. But now I updated it from the website itself.

src/lib/free-queue/free-queue-sab.js
```diff
@@ -231,4 +231,5 @@ class FreeQueueSAB {
     }
   }
   
-  export { FreeQueueSAB };
\ No newline at end of file
+  // export { FreeQueueSAB };
```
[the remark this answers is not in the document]
professorabhay: No, That was for testing purposes.
I updated it back to its original state.

src/lib/free-queue/coi-serviceworker.js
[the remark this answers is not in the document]
professorabhay: Yes, that's what I thought. We will look into this later on. For now, having a separate file is a better approach.

professorabhay:
image


http://localhost:8080/lib/free-queue/test.html

src/lib/free-queue/coi-serviceworker.js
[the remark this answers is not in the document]
mjwilson-google: That sounds good.

src/lib/free-queue/test.html
mjwilson-google: Thank you, it's working for me now.

src/lib/free-queue/test.html
mjwilson-google: The tests are running now for me, but I still see the SyntaxError in the console.  It's in chai.js so maybe there's not much we can do?

mjwilson-google:
I'm seeing the same failure in the "should fail to push data when buffer is full" test.  I'll take a look too.

professorabhay:
> I'm seeing the same failure in the "should fail to push data when buffer is full" test.  I'll take a look too.

I figured it out. I'll push the update tomorrow.

.eleventy.js
```diff
@@ -79,6 +79,8 @@ module.exports = function(eleventyConfig) {
     'src/README.md',
     'src/sitemap.xml',
     'src/lib/**/*.js',
+    'src/lib/**/*.html',
+    'src/lib/**/**/*.js',
```
mjwilson-google, line 83: I think this is redundant with `'src/lib/**/*.js',` -- the `**` should match any number of subdirectrories.

professorabhay: Oh, I see. 
Thanks for pointing out. 
I'll update it.

src/lib/free-queue/test/free-queue-sab.test.js
```diff
@@ -0,0 +1,113 @@
...
+            expect(result).to.be.true;
+            expect(queue.getAvailableSamples()).to.equal(3);
+        });
+
+        it('should fail to push data when buffer is full', function() {
+            const input = [new Float32Array(queue.getBufferLength()), new Float32Array(queue.getBufferLength())];
+            
+            let result = queue.push(input, queue.getBufferLength());
+            console.log("Push result when buffer is full:", result);
+            expect(result).to.be.true;
```
mjwilson-google, line 48: Instead of logging to the console here, let's use the message feature of the `expect()` function:
`expect(result, 'Push result when buffer is full').to.be.true;`

src/lib/free-queue/test/free-queue-sab.test.js
```diff
@@ -0,0 +1,113 @@
...
+    
+            result = queue.push(input, queue.getBufferLength());
+            console.log("Push result when attempting to overfill buffer:", result);
+            expect(result).to.be.false;
```
mjwilson-google, line 52: Same here, remove the console logging and add it as a message:
`expect(result, 'Push result when attempting to overfill buffer').to.be.false;`

src/lib/free-queue/test/free-queue-sab.test.js
```diff
@@ -0,0 +1,113 @@
...
+
+        it('should correctly check if frame is available', function() {
+            expect(queue.isFrameAvailable(3)).to.be.false;
+            const input = [new Float32Array([1, 2, 3]), new Float32Array([4, 5, 6])];
+            queue.push(input, 3);
+            expect(queue.isFrameAvailable(3)).to.be.true;
+        });
+    });
+
+    describe('Performance Tests', function() {
```
mjwilson-google, line 90: It looks like Mocha will display timing information for "slow" tests.  This isn't very accurate, but may be enough for our first pass.  I recommend removing the console logging and timeout, and rewriting this section to just have two tests that repeatedly push and pull data, one large and one small.

I tried to modify the existing tests and came up with this which seemed to work well; please feel free to use it or modify as you like.  You may have to adjust the number of iterations to get the timing to show up.

```
  describe('Performance Tests', function() {
    it('should handle large data efficiently', function() {
      const input = [new Float32Array(bufferLength), new Float32Array(bufferLength)];
      const output = [new Float32Array(bufferLength), new Float32Array(bufferLength)];
      for (let i = 0; i < 100000; i++) {
        queue.push(input, bufferLength);
        queue.pull(output, bufferLength);
      }
    });

    it('should handle small data efficiently', function() {
      const input = [new Float32Array(1), new Float32Array(1)];
      const output = [new Float32Array(1), new Float32Array(1)];
      for (let i = 0; i < 1000000; i++) {
        queue.push(input, 1);
        queue.pull(output, 1);
      }
    });
  });
```

src/lib/free-queue/test/free-queue-sab.test.js
```diff
@@ -0,0 +1,113 @@
...
+            queue.push(largeInput, bufferLength);
+            console.timeEnd('push large data');
+
+            const largeOutput = [new Float32Array(bufferLength), new Float32Array(bufferLength)];
+            console.time('pull large data');
+            queue.pull(largeOutput, bufferLength);
+            console.timeEnd('pull large data');
+        });
+
+        it('should handle maximum capacity', function() {
```
mjwilson-google, line 104: This test should be moved to the `push` section above.

src/lib/free-queue/test/free-queue-sab.test.js
```diff
@@ -0,0 +1,113 @@
...
+            this.timeout(10000);
```
mjwilson-google, line 105: Remove the timeout, it's not necessary for this test.

mjwilson-google (changes requested):
Looks good overall.  If you agree with my other comments please make the changes and we can merge this.

I noticed that there is still the console error about the unexpected export token in chai.js, but it doesn't seem to cause any problems so I don't think we need to worry about it.

professorabhay:
We are good to merge this PR now. 
That error on the console is comping because of the webpack bundler but it's not impacting anything in our tests. 
I'll try to find any way to resolve this and if find then update it in the new PR.

src/lib/free-queue/test/free-queue-sab.test.js
```diff
@@ -0,0 +1,131 @@
...
+
+        it('should correctly check if frame is available', function () {
+            expect(queue.isFrameAvailable(3)).to.be.false;
+            const input = [new Float32Array([1, 2, 3]), new Float32Array([4, 5, 6])];
+            queue.push(input, 3);
+            expect(queue.isFrameAvailable(3)).to.be.true;
+        });
+    });
+
+    // describe('Performance Tests', function() {
```
mjwilson-google, line 96: Sorry, one last change: please just remove this code instead of commenting it out.  Everything else looks good.

mjwilson-google (changes requested):

src/lib/free-queue/test/free-queue-sab.test.js
professorabhay: Ah, My bad!

mjwilson-google (approved):

PR #385, opened 2024-07-20
Add style and realtime sine reference to live suite build

### LINKED ISSUE #382: gh-pages deployment doesn't build non-HTML/JS files.
The [live suite](https://googlechromelabs.github.io/web-audio-samples/tests/playwright/pages/) fetches a `live-suite/style.css` and a reference file `reference/440@48k-sine-octx.json`. While these files exist on the `main` branch, `gh-pages` doesn't have the same files: https://github.com/GoogleChromeLabs/web-audio-samples/tree/c29cec74c47e58b95def50eb8c166bcfb88e45e7/tests/playwright/pages. This causes the live suite to break when not run locally.

Kizjkre:
Fix #382.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,6 +24,9 @@ jobs:
       run: |
         npm install
         npm run build
+        cp src/tests/playwright/pages/live-suite/style.css _site/tests/playwright/pages/live-suite/style.css
+        mkdir _site/tests/playwright/pages/reference
+        cp src/tests/playwright/pages/reference/440@48k-sine-octx.json _site/tests/playwright/pages/reference/440@48k-sine-octx.json
     - name: deploy _site to gh-pages
       uses: s0/git-publish-subdir-action@develop
       env:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,9 +24,6 @@ jobs:
       run: |
         npm install
         npm run build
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -24,6 +24,9 @@ jobs:
       run: |
         npm install
         npm run build
+        cp src/tests/playwright/pages/live-suite/style.css _site/tests/playwright/pages/live-suite/style.css
```
mjwilson-google, line 27: Could we add  `src/tests/playwright/pages/live-suite/style.css` to [.eleventy.js](https://github.com/GoogleChromeLabs/web-audio-samples/blob/main/.eleventy.js#L46) instead?

.github/workflows/main.yml
```diff
@@ -24,6 +24,9 @@ jobs:
...
+        mkdir _site/tests/playwright/pages/reference
+        cp src/tests/playwright/pages/reference/440@48k-sine-octx.json _site/tests/playwright/pages/reference/440@48k-sine-octx.json
```
mjwilson-google, line 29: Similar, could we add  `src/tests/playwright/pages/reference/*.json` to [.eleventy.js](https://github.com/GoogleChromeLabs/web-audio-samples/blob/main/.eleventy.js#L46) instead?

mjwilson-google (changes requested):

mjwilson-google (approved):

PR #388, opened 2024-07-23
FreeQueue: automated test coverage

professorabhay:
- Test cases for FreeQueue Class 

Fixes https://github.com/GoogleChromeLabs/web-audio-samples/issues/347

### DIFF
--- a/src/lib/free-queue/free-queue.js
+++ b/src/lib/free-queue/free-queue.js
@@ -191,7 +191,7 @@ class FreeQueue {
     // match with this buffer obejct.
 
     if (arraySequence.length !== this._channelCount) {
-      throw new Error('Channel count mismatch');
+      throw new Error(`Channel count mismatch: expected ${this._channelCount}, but got ${arraySequence.length}.`);
     }
 
     // Transfer data from the |arraySequence| storage to the internal buffer.
@@ -220,7 +220,7 @@ class FreeQueue {
     // match with this buffer obejct.
 
     if (arraySequence.length !== this._channelCount) {
-      throw new Error('Channel count mismatch');
+      throw new Error(`Channel count mismatch: expected ${this._channelCount}, but got ${arraySequence.length}.`);
     }
 
     // If the FIFO is completely empty, do nothing.
--- a/src/lib/free-queue/free-queue.js
+++ b/src/lib/free-queue/free-queue.js
@@ -190,6 +190,10 @@ class FreeQueue {
     // The channel count of arraySequence and the length of each channel must
     // match with this buffer obejct.
 
+    if (arraySequence.length !== this._channelCount) {
+      throw new Error('Channel count mismatch');
+    }
+
     // Transfer data from the |arraySequence| storage to the internal buffer.
     const sourceLength = arraySequence[0].length;
     for (let i = 0; i < sourceLength; ++i) {
@@ -214,6 +218,10 @@ class FreeQueue {
   pull(arraySequence) {
     // The channel count of arraySequence and the length of each channel must
     // match with this buffer obejct.
+
+    if (arraySequence.length !== this._channelCount) {
+      throw new Error('Channel count mismatch');
+    }
 
     // If the FIFO is completely empty, do nothing.
     if (this._framesAvailable === 0) {
--- a/src/lib/free-queue/test/free-queue.test.html
+++ b/src/lib/free-queue/test/free-queue.test.html
@@ -1,38 +1,49 @@
 <!DOCTYPE html>
 <html lang="en">
+
 <head>
-    <meta charset="UTF-8">
-    <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>FreeQueue Tests</title>
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/mocha/10.6.0/mocha.css" integrity="sha512-xPbP9qFwQklQFQX6R6+36vNq6mZatPrpfEKUB/zWASwZvfDBy8Y2gEpkRuDlZ0WCiZZGMPyluZif5v2KKxylqg==" crossorigin="anonymous" referrerpolicy="no-referrer" />
-    <style>
-      .head {
-          margin-top: 5vh;
-          text-align: center;
-      }
-      button {
-            margin-left: 45vw;
-            margin-top: 30vh;
-            padding: 10px 20px;
-            font-size: 16px;
-        }
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>FreeQueue Tests</title>
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/mocha/10.6.0/mocha.css"
+    integrity="sha512-xPbP9qFwQklQFQX6R6+36vNq6mZatPrpfEKUB/zWASwZvfDBy8Y2gEpkRuDlZ0WCiZZGMPyluZif5v2KKxylqg=="
+    crossorigin="anonymous" referrerpolicy="no-referrer" />
+  <style>
+    .head {
+      margin-top: 5vh;
+      text-align: center;
+    }
+
+    button {
+      margin-left: 45vw;
+      margin-top: 30vh;
+      padding: 10px 20px;
+      font-size: 16px;
+    }
   </style>
 </head>
+
 <body>
-    <h1 class="head">FreeQueue Test</h1>
-    <div id="mocha"></div>
-    <button id="run-tests">Run Tests</button>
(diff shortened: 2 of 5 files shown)


src/lib/free-queue/test/free-queue.test.js
```diff
@@ -0,0 +1,147 @@
...
+
+    it('should throw an error if pulling with mismatched channel count', () => {
+      const invalidOutputData = [new Float32Array(bufferLength)];
+      expect(() => freeQueue.pull(invalidOutputData)).to.throw(Error, 'Channel count mismatch');
+    });
+  });
+
+  describe('Performance', () => {
+    it('should efficiently handle large data transfers', function() {
+      this.timeout(5000);
```
hoch, line 120: Can we avoid using `timeout`?

src/lib/free-queue/free-queue.js
```diff
@@ -190,6 +190,10 @@ class FreeQueue {
     // The channel count of arraySequence and the length of each channel must
     // match with this buffer obejct.
 
+    if (arraySequence.length !== this._channelCount) {
+      throw new Error('Channel count mismatch');
```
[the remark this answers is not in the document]
professorabhay: Ok, I'll look into that.

src/lib/free-queue/test/free-queue.test.html
```diff
@@ -0,0 +1,38 @@
...
+            font-size: 16px;
+        }
+  </style>
+</head>
+<body>
+    <h1 class="head">FreeQueue Test</h1>
+    <div id="mocha"></div>
+    <button id="run-tests">Run Tests</button>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/mocha/10.6.0/mocha.min.js" integrity="sha512-Wt8lK9Rbdq3uAOR9lm7Fy+XTPP0xie3DzFQ0JTzQLGPvdZ7tDFMqHOJq6hbB4v8uvj7xaBEsINRs7K7HTfi9wA==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/chai/5.1.1/chai.js" integrity="sha512-HLs4bXIGAstDHJcCm2HfGhBkTUW9oErgCOCiDEnoOhwCEJLcLZUXX7TKcarluZx1D385vuUziYjD1uNGXy9zzg==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
```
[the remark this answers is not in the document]
professorabhay: As far I know chai doesn't have anything like that!

src/lib/free-queue/test/free-queue.test.js
professorabhay: Let me see!

mjwilson-google:
Let's merge #374 first, then make sure this class has the same amount of coverage.

src/lib/free-queue/test/free-queue.test.js
```diff
@@ -0,0 +1,162 @@
...
+  // Simulate memory allocation
+  _malloc: (size) => new ArrayBuffer(size), 
+  // Simulate memory deallocation
+  _free: () => {}, 
+  // Simulate HEAPF32
+  HEAPF32: new Float32Array(1024), 
+};
+
+describe('FreeQueue Class', () => {
+  const bufferLength = 512;
```
mjwilson-google, line 15: Let's change this to 1024 to match the FreeQueueSAB tests, so that the performance tests can be compared more directly.

src/lib/free-queue/test/free-queue.test.js
```diff
@@ -0,0 +1,162 @@
...
+      
+      const expectedChannelCount = freeQueue._channelCount; 
+      const actualChannelCount = invalidOutputData.length;
+    
+      expect(() => freeQueue.pull(invalidOutputData))
+        .to.throw(Error, `Channel count mismatch: expected ${expectedChannelCount}, but got ${actualChannelCount}.`);
+    });
+  });
+
+  describe('Performance', () => {
```
mjwilson-google, line 131: Let's remove these and port the large and small data tests from FreeQueueSAB tests here instead.

src/lib/free-queue/test/free-queue.test.js
```diff
@@ -0,0 +1,162 @@
...
+    });
+
+    it('should allocate the correct amount of memory', () => {
+      const dataByteSize = channelCount * bufferLength * Float32Array.BYTES_PER_ELEMENT;
+      expect(freeQueue.getPointer()).to.be.instanceof(ArrayBuffer);
+      expect(freeQueue.getPointer().byteLength).to.equal(dataByteSize);
+    });
+  });
+
+  describe('Channel Adaptation', () => {
```
mjwilson-google, line 42: Can we port this test to the FreeQueueSAB test suite as well, or does that not allow a variable channel count?

Can we also add another test that tries maxChannelCount+1 and shows the channel count doesn't update / an error is thrown / etc.?

src/lib/free-queue/test/free-queue.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+        expect(outputData[0]).to.deep.equal(testData[0]);
+        expect(outputData[1]).to.deep.equal(testData[1]);
+        expect(freeQueue.framesAvailable).to.equal(0);
+      }
+    });
+  });
+
+  describe('Error Handling', () => {
+    it('should return null for invalid channel index in getChannelData', () => {
```
mjwilson-google, line 105: Can we port these tests back to the FreeQueueSAB suite too?

src/lib/free-queue/test/free-queue.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+  describe('Channel Adaptation', () => {
+    it('should adapt to a new channel count within limits', () => {
+      freeQueue.adaptChannel(3);
+      expect(freeQueue.numberOfChannels).to.equal(3);
+    });
+  });
+
+  describe('Pushing and Pulling Data', () => {
+    it('should correctly push and pull data', () => {
```
mjwilson-google, line 50: Let's match these tests with the 'push' and 'pull' sections from the FreeQueueSAB suite.

src/lib/free-queue/test/free-queue.test.js
```diff
@@ -0,0 +1,162 @@
...
+      const invalidOutputData = [new Float32Array(bufferLength)]; 
+      
+      const expectedChannelCount = freeQueue._channelCount; 
+      const actualChannelCount = invalidOutputData.length;
+    
+      expect(() => freeQueue.pull(invalidOutputData))
+        .to.throw(Error, `Channel count mismatch: expected ${expectedChannelCount}, but got ${actualChannelCount}.`);
+    });
+  });
+
```
mjwilson-google, line 130: Can we port the "Utility Methods" tests from the FreeQueueSAB tests here too?  Or is the API different?

mjwilson-google (changes requested):
Now that #374 is landed, let's take the best aspects of both PRs and apply them to both test suites.

src/lib/free-queue/test/free-queue.test.js
professorabhay: Should we just define 2 sperate describe blocks for push and pull with the same test cases?

src/lib/free-queue/test/free-queue.test.js
professorabhay: I thing we can't do this. Because the functions used in the "Utility Methods" are not available in the FreeQueue Class.

src/lib/free-queue/test/free-queue.test.js
professorabhay: This is possible to do. I'll do this in a new PR after merging this.

src/lib/free-queue/test/free-queue.test.js
professorabhay: I'll look into it and let you know!

src/lib/free-queue/test/free-queue.test.js
mjwilson-google: Yes, I think when the APIs for the classes are the same we should use the same tests and format.  So since we did this in two blocks for SAB, I think we should do it in two blocks here too.

professorabhay: Ok, I'll update this!

mjwilson-google (approved):
Looks good to me!  Let's follow-up with another PR later to make both test suites more similar.

When I ran this., FreeQueue was significantly faster than FreeQueueSAB (about 2x).  Is that what you were seeing too?  We have to be a bit careful about reading too much into the results because this isn't a very accurate timing method though.

Please merge when you are ready.

professorabhay:
Yeah Sure!

Hmm, I am not sure but yes there was some difference. I'll look into that.

Sure, we can merge this PR and I'll raise a new PR for extending the SAB TestCases and resolve some nits.

mjwilson-google:
Sorry, didn't realize you weren't authorized!

PR #391, opened 2024-08-16
Added more test cases - ChannelAdaption and ErrorHandling

professorabhay:
Moved some test cases from the `FreeQueue Test Suite` and extended the test suite to cover more areas.


mjwilson-google (approved):
Looks good, thank you.  Would be nice to match the names and order of the tests across both suites, but that is a minor point and can be done in a different PR.

PR #399, opened 2024-08-27
minor change- spelling correction

professorabhay:


mjwilson-google (approved):
This is actually changing the class type!  Good catch.

PR #403, opened 2024-10-01
Add .nojekyll file, resolves #402

### LINKED ISSUE #402: GitHub Pages removes the rainfly/_app/ directory
# To reproduce:
1. Visit [Rainfly](https://googlechromelabs.github.io/web-audio-samples/rainfly/)
2. Open the JavaScript console

[This thread](https://github.com/orgs/community/discussions/61073#discussioncomment-7016026) talks about how GitHub processes GitHub Pages through Jekyll, which removes directories that start with `_`. Rainfly uses the `_app` directory for much of the JS files, which gets removed by GH Pages, giving a 404 when trying to access them.

# To fix:
Add a `.nojekyll` file to the root directory in the `gh-pages` branch.

Kizjkre:


mjwilson-google (commented):
Would it also be possible to fix this by renaming the `_app` directory to something that doesn't start with an underscore?

mjwilson-google (approved):
Got it, thanks!

terryzfeng:
> Would it also be possible to fix this by renaming the `_app` directory to something that doesn't start with an underscore?

I managed to rename _app to app in the svelte build pipeline. Thanks for the suggestion @mjwilson-google !

PR #407, opened 2025-02-01
Fix some typos;

eof273:


mjwilson-google:
As noted by the google-cla bot above, we need a signed CLA before we can merge the change.

Thank you for taking the time to find and fix all of these typos!

hoch:
Since this PR became stale, I think we need to duplicate the contents and create a new PR.

PR #412, opened 2025-03-24
Resampler Smoke Test (Part 1)

hoch:
- Adding foundation for A/B testing
- Simple logging for sample rate, base latency, and computed buffer size

### DIFF
--- a/src/_data/build_info.json
+++ b/src/_data/build_info.json
@@ -1,1 +1,1 @@
-{"version":"3.2.1","revision":"df05f31b","lastUpdated":"2025-03-24","copyrightYear":2025}
+{"version":"3.2.1","revision":"eefc66ff","lastUpdated":"2025-03-24","copyrightYear":2025}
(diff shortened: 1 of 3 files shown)


alvinjiooo:
When test this locally, I found that the link to resampler and resampler-smoke are not added in src/tests/index.njk 
But if I directly paste the url (http://localhost:8080/tests/resampler-smoke/) works fine. 
I wonder should we add the link to "Test" page?

alvinjiooo:
> When test this locally, I found that the link to resampler and resampler-smoke are not added in src/tests/index.njk But if I directly paste the url (http://localhost:8080/tests/resampler-smoke/) works fine. I wonder should we add the link to "Test" page?

Other than this the change LGTM, I tested this locally on engineering build which has AudioDestination resampler removed and result is same as formal build with the resampler removal change.

hoch:
> When test this locally, I found that the link to resampler and resampler-smoke are not added in src/tests/index.njk

This is expected. It is necessary to add links manually when the test file is ready. I am treating this test app as a development build. So we're not exposing it to the index page yet.

PR #415, opened 2025-04-15
Add listener to wavetable synth for autoplay

mjwilson-google:
The wavetable synth will not make sound if it is navigated to directly, because there isn't a user gesture so the autoplay rules will be enforced.

This PR adds listeners to allow the user to click or press a key to start playing, even if the page is loaded directly.

### DIFF
--- a/src/demos/wavetable-synth/index.html
+++ b/src/demos/wavetable-synth/index.html
@@ -955,15 +955,29 @@ function addUI() {
 }
 
 function start() {
-    // first, get rid of loading animation
     var loading = document.getElementById("loading");
-    loading.innerHTML = "";
 
-    startTime = context.currentTime + 0.160;
-    
     waveTable = loader.getTable(defaultWaveTableName);
     waveTable2 = loader.getTable(defaultWaveTableName);
-    sequence.schedule();    
+
+    const startAudioPlayback = () => {
+        loading.innerHTML = "";
+        startTime = context.currentTime + 0.160;
+        sequence.schedule();
+    };
+
+    if (context.state === 'running') {
+        startAudioPlayback();
+    } else if (context.state === 'suspended') {
+        loading.innerHTML = "Click anywhere or press a key to start playing.";
+        const interactionHandler = () => {
+            document.removeEventListener('mousedown', interactionHandler);
+            document.removeEventListener('keydown', interactionHandler);
+            context.resume().then(() => startAudioPlayback());
+        };
+        document.addEventListener('mousedown', interactionHandler);
+        document.addEventListener('keydown', interactionHandler);
+    }
 }
 
 </script>


src/demos/wavetable-synth/index.html
```diff
@@ -955,15 +955,29 @@
...
+        sequence.schedule();
+    };
+
+    if (context.state === 'running') {
+        startAudioPlayback();
+    } else if (context.state === 'suspended') {
+        loading.innerHTML = "Click anywhere or press a key to start playing.";
+        const interactionHandler = () => {
+            document.removeEventListener('mousedown', interactionHandler);
+            document.removeEventListener('keydown', interactionHandler);
```
hoch, line 975: Does this work correctly? I didn't know a key stroke is treated as a user gesture.

src/demos/wavetable-synth/index.html
```diff
@@ -955,15 +955,29 @@
...
+        startTime = context.currentTime + 0.160;
+        sequence.schedule();
+    };
+
+    if (context.state === 'running') {
+        startAudioPlayback();
+    } else if (context.state === 'suspended') {
+        loading.innerHTML = "Click anywhere or press a key to start playing.";
+        const interactionHandler = () => {
+            document.removeEventListener('mousedown', interactionHandler);
```
hoch, line 974: You can remove this.

src/demos/wavetable-synth/index.html
```diff
@@ -955,15 +955,29 @@
...
+            document.removeEventListener('keydown', interactionHandler);
+            context.resume().then(() => startAudioPlayback());
+        };
+        document.addEventListener('mousedown', interactionHandler);
```
hoch, line 978: Instead, you can do this:
```js
document.addEventListener('mousedown', interactionHandler, {once: true});
```

src/demos/wavetable-synth/index.html
```diff
@@ -955,15 +955,29 @@
...
+        document.addEventListener('keydown', interactionHandler);
```
hoch, line 979: Ditto

hoch (commented):
LGTM with nits

hoch (approved):
(Approving)

src/demos/wavetable-synth/index.html
mjwilson-google: It seems to work when I tried it!

src/demos/wavetable-synth/index.html
mjwilson-google: It's one less line overall to check the state in `interactionHandler`, do you think that would be cleaner?  I don't have a strong preference.

mjwilson-google: I'll land as-is and we can modify later if necessary.  Thank you for the review!

PR #420, opened 2025-06-06
clarifying statement

lourendotco:
there seems to be some confusion in the previous comment.

### DIFF
--- a/src/lib/free-queue/free-queue.js
+++ b/src/lib/free-queue/free-queue.js
@@ -17,9 +17,6 @@
 // Byte per audio sample. (32 bit float)
 const BYTES_PER_SAMPLE = Float32Array.BYTES_PER_ELEMENT;
 
-// Basic byte unit of WASM heap. (16 bit = 2 bytes)
-const BYTES_PER_UNIT = Uint16Array.BYTES_PER_ELEMENT;
-
 // The max audio channel on Chrome is 32.
 const MAX_CHANNEL_COUNT = 32;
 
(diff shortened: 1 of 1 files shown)


src/lib/free-queue/free-queue.js
```diff
@@ -80,11 +77,11 @@ class FreeQueue {
...
       const startByteOffset = this._dataPtr + i * channelByteSize;
       const endByteOffset = startByteOffset + channelByteSize;
-      // Get the actual array index by dividing the byte offset by 2 bytes.
+      // Get the actual array index by dividing the byte offset by 4, the size of float. (right shift >> 2 === dividing by 4)
       this._channelData[i] =
           this._module.HEAPF32.subarray(
-              startByteOffset >> BYTES_PER_UNIT,
-              endByteOffset >> BYTES_PER_UNIT);
+              startByteOffset >> 2,
+              endByteOffset >> 2);
```
mjwilson-google, line 84: For even more clarity, what do you think about either of these options?

1. Make the `2` a named constant like `FLOAT32_INDEX_SHIFT`

or

2. Use division by `BYTES_PER_SAMPLE` along with `Math.floor()`

src/lib/free-queue/free-queue.js
```diff
@@ -249,4 +246,4 @@ export {
   MAX_CHANNEL_COUNT,
   RENDER_QUANTUM_FRAMES,
   FreeQueue
-};
\ No newline at end of file
+};
```
mjwilson-google, line 249: Nit: is it possible to keep this unchanged (that is, don't add an additional newline)?

mjwilson-google (commented):
Thank you for your contribution!  I think you're right, and this is much more clear.

I have a couple of nits/questions; please let me know what you think.

PR #457, opened 2026-03-23
Update dependencies and migrate configurations (ESLint v10, Tailwind v4, Eleventy 3)

hoch:
This PR addresses several major package upgrades and resolves multiple underlying security vulnerabilities flagged by Dependabot. Since these dependency bumps involved major versions, several configuration architectures have been updated to reflect the new tooling standard.

### Changes Made
* **Dependency Updates**: Flushed and regenerated the `package-lock.json` and updated root dependencies to their latest compatible versions. 
* **ESLint v10 Migration**: 
  * Removed the deprecated `.eslintrc.json`.
  * Dropped the obsolete `eslint-config-google` in favor of standard `@eslint/js` recommended rules.
  * Implemented the newly required Flat Config format via `eslint.config.mjs`.
* **Tailwind CSS v4 Migration**: 
  * Adopted the new CSS-first architectural pattern (`@import "tailwindcss";` and `@source`).
  * Deleted obsolete `tailwind.config.js` and `postcss.config.js`.
* **Eleventy v3 Compliance**: 
  * Updated the deprecated `.eleventy.js` dev server configuration mapping, replacing `eleventyConfig.setBrowserSyncConfig()` with the native `eleventyConfig.setServerOptions()`.

### Security Resolutions
Regenerating the lockfile successfully resolves the following associated Dependabot alerts:
* Closes #456 (flatted CWE-1321)
* Closes #454 (liquidjs path traversal)
* Closes #453 (minimatch ReDoS vulnerability)
* Closes #450 (ajv $data exploit)
* Closes #444 (markdown-it security regression)

### Testing Instructions
1. Pull the branch locally.
2. Run `npm install` (and `npm install -D globals` if needed for linting).
3. Verify the site renders identically by running `npm run start` (or `npm run build`).
4. Ensure the linting hooks run error-free with `npm run format:eslint`.

*** 

Verification

local_site_verification_1774291494266


eslint.config.mjs
```diff
@@ -0,0 +1,31 @@
...
+    languageOptions: {
+      ecmaVersion: 2021,
+      sourceType: "module",
+      globals: {
+        ...globals.browser
+      }
+    },
+    rules: {
+      "require-jsdoc": "off",
+      // Note: We dropped eslint-config-google because it is deprecated
```
mjwilson-google, line 27: It looks like this config doesn't enforce any style; if that's the case should we just remove eslint completely?

.npmrc
```diff
@@ -0,0 +1 @@
+registry=https://registry.npmjs.org/
```
mjwilson-google, line 1: I think this is also the default, correct?

mjwilson-google (approved):
LGTM with some comments

.npmrc
hoch: Yes, but adding this in .npmrc is to instruct the npm always use the public registry. It prevents `package-lock` from having potential internal package URLs. (a practice recommended by Gemini)

eslint.config.mjs
hoch: It looks like we don't have the official eslint support for JS, but still having eslint is necessary for enforcing the basic JS linting for any PRs moving forward.

hoch: I'll need to investigate the roadmap or recommended practice for linting. This PR purely focuses on the upgrading packages to resolve problematic depedencies.

Issue #473, opened 2026-08-22
Several AudioWorklet demos have no way to stop playback except reloading the page


Aditya30ag:
### Description
A number of demos under `src/audio-worklet/` only implement a **START** button. Once clicked, playback begins and the button disables itself there is no Stop control. The only way to stop the audio is to reload the page.

### Affected demos
- `src/audio-worklet/basic/handling-errors/`
- `src/audio-worklet/basic/message-port/`
- `src/audio-worklet/design-pattern/shared-buffer/`
- `src/audio-worklet/design-pattern/wasm-ring-buffer/`
- `src/audio-worklet/design-pattern/wasm/`

### Steps to reproduce
1. Open any of the demos above.
2. Click **START**.
3. Try to stop playback no control is available.

### Expected behavior
A visible **STOP** button should let the user suspend the audio
context and re-enable Start, without reloading the page.

### Proposed fix
Add a `button-stop` element next to `button-start` in each demo's markup, and wire it to `audioContext.suspend()`, re-enabling Start when clicked. Guard worklet-module loading and graph setup with flags so Start doesn't rebuild the audio graph on repeat clicks.

PR #474, opened 2026-08-24
Add Stop button to AudioWorklet demos missing one

Aditya30ag:
Several AudioWorklet demos only provide a **START** button, with no way to stop playback without reloading the page.

Add a visible **STOP** button to the affected demos. The button should suspend the audio context and re-enable **START**. Also guard worklet loading and graph setup so repeated starts do not rebuild the audio graph.

## Before -
image

## After -
image
Issue-#473

### DIFF
--- a/src/audio-worklet/basic/handling-errors/index.njk
+++ b/src/audio-worklet/basic/handling-errors/index.njk
@@ -20,6 +20,7 @@ to_root_dir: ../../../
 
 <div class="demo-box">
   <button id="button-start" disabled>START</button>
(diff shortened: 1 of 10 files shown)


Aditya30ag:
@hoch please have a look

hoch:
Thanks for your work on this, @Aditya30ag! We are currently working on a redesign of the site. We’ll keep this PR in mind and revisit implementing the stop-button functionality once the new structure is in place.

Aditya30ag:
Thanks for the update! Sounds good.

